PR #6, opened 2019-05-29
WINDUP-2392 Deploy to Insights CI Environment

mrizzi:
https://issues.jboss.org/browse/WINDUP-2392
Set of basic changes to have the prototype to work in Insights CI environment:

- added some routes' ID to help in debugging (auto-generated `route#` IDs where not so useful)
- added Kafka route to log the validation messages
- added `@JsonIgnoreProperties` annotation to `RHIdentity` in order to avoid the exception `com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "user" (class org.jboss.xavier.integrations.route.model.RHIdentity), not marked as ignorable (2 known properties: "account_number", "internal"])`
- added a new filter approach for kafka messages to avoid a NPE during base64 decoding of `b64_identity` value when it's missing (it happend with a message from `service=compliance`)
- added/changed some properties
- added a new `analytics_template_insights_ci.json` OCP template
- added some labels to have the delete command to work using label
- changed accordingly the README

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
+++ b/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
@@ -133,6 +133,7 @@ public class MainRouteBuilder extends RouteBuilder {
                     }
                 })
                 .unmarshal().json(JsonLibrary.Jackson, FilePersistedNotification.class)
+                .filter(simple("'ma-xavier' == ${body.getCategory}"))
                 .to("direct:download-from-S3");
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -130,14 +133,18 @@ public void configure() {
                     }
                 })
                 .unmarshal().json(JsonLibrary.Jackson, FilePersistedNotification.class)
+                .filter(simple("'ma-xavier' == ${body.getCategory}"))
```
jonathanvila, line 136: should we put this into a Property ?

PR #12, opened 2019-07-03
WINDUP-2392 Prototype in CI env

mrizzi:
https://issues.jboss.org/browse/WINDUP-2392

- applied the changes needed to have the integration to work with the new Insights upload service
- fixed the `x-rh-insights-request-id` to be dynamic
- added the `DefaultHeaderFilterStrategy` to our REST configuration to avoid the copy of the request's header into the response's header


mrizzi:
I know it's not the standard way but i'm merging this PR without review in order to have this code to be available on master branch ASAP to try to promote the prototype in QA.

PR #17, opened 2019-07-23
MIGENG-130 VMWorkloadInventoryModel

m-brophy:
There are comments (which could be removed if it's felt they're inappropriate) showing the intended usage of fields.
Also I've provided setX and addX methods for fields which will be populated via traversal of 0-many input payload objects. they can be populated all in one go using the setX methods or incrementally as we traverse using addX methods, depending on how we want to do it. We can delete whichever population methods we don't need later on if necessary.

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
@@ -3,6 +3,7 @@ package org.jboss.xavier.analytics.pojo.input;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 
 
@@ -33,12 +34,12 @@ public class VMWorkloadInventoryModel
     //hardware/disks/filename
     private Collection<String> vmDiskFilenames;
     private Collection<String> systemServicesNames;
-    private Collection<Map<String,String>> files;
+    private Map<String,String> files;
 
     public VMWorkloadInventoryModel()
     {
         this.systemServicesNames = new ArrayList<> ();
-        this.files = new ArrayList<>();
+        this.files = new HashMap<>();
         this.vmDiskFilenames = new ArrayList<>();
         nicsCount = 0;
         diskSpace = new BigDecimal(0);
@@ -164,16 +165,16 @@ public class VMWorkloadInventoryModel
         this.systemServicesNames.add(systemServiceName);
     }
 
-    public Collection<Map<String, String>> getFiles() {
+    public Map<String, String> getFiles() {
         return files;
     }
 
 
-    public void setFiles(Collection<Map<String, String>> files) {
+    public void setFiles(Map<String, String> files) {
         this.files = files;
     }
 
-    public void addFile(Map<String, String> file) {
-        this.files.add(file);
+    public void addFile(String name, String contents) {
+        this.files.put(name,contents);
     }
 }
--- a/src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
@@ -1,0 +1,179 @@
+package org.jboss.xavier.analytics.pojo.input;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Map;
+
+
+public class VMWorkloadInventoryModel
+{
+    //common/name
+    private String provider;
+    //common/ems_clusters/v_parent_datacenter
+    private String datacenter;
+    //common/ems_clusters/name
+    private String cluster;
+    //vms/name
+    private String vmName;
+    //sum of vms/hardware/disks/size_on_disk
+    private BigDecimal diskSpace;
+    //vms/ram_size_in_bytes
+    private Integer memory;
+    //vms/num_cpu
+    private Integer cpuCores;
+    //vms/operating_system/product_name
+    private String osProductName;
+    //hardware/guest_os_full_name
+    private String guestOSFullName;
+    //vms/has_rdm_disk
+    private boolean hasRdmDisk;
+    //count of nics object within the vms/hardware
+    private Integer nicsCount;
+    //hardware/disks/filename
+    private Collection<String> vmDiskFilenames;
+    private Collection<Collection<Map<String, String>>> systemServices;
+    private Collection<Collection<Map<String,String>>> files;
+
+    public VMWorkloadInventoryModel()
+    {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+    private String osProductName;
+    //hardware/guest_os_full_name
+    private String guestOSFullName;
+    //vms/has_rdm_disk
+    private boolean hasRdmDisk;
+    //count of nics object within the vms/hardware
+    private Integer nicsCount;
+    //hardware/disks/filename
+    private Collection<String> vmDiskFilenames;
+    private Collection<Collection<Map<String, String>>> systemServices;
```
jonathanvila: This is the structure : 
],
            "system_services": [
              {
                "id": 9669,
                "name": "rhnsd",
                "typename": "linux_initprocess"
              },
So unless it's dynamic I would suggest having a Class representing the fields

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+    private Collection<Collection<Map<String,String>>> files;
```
jonathanvila: In my example payload I dont have detail on **files** , but I assume we could have the same as with services

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+    //vms/operating_system/product_name
+    private String osProductName;
+    //hardware/guest_os_full_name
+    private String guestOSFullName;
+    //vms/has_rdm_disk
+    private boolean hasRdmDisk;
+    //count of nics object within the vms/hardware
+    private Integer nicsCount;
+    //hardware/disks/filename
+    private Collection<String> vmDiskFilenames;
```
jonathanvila, line 35: If we are gonna have the list of the names of the disks, we could also have their size, and not only the sum on **diskSpace**
What would be the use of having only the list of harddisk names ?

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.Collection;
+import java.util.Map;
+
+
+public class VMWorkloadInventoryModel
+{
+    //common/name
+    private String provider;
+    //common/ems_clusters/v_parent_datacenter
+    private String datacenter;
```
jonathanvila: Which one of the ems_clusters would be referred ?

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+    //common/ems_clusters/name
+    private String cluster;
```
jonathanvila: which one of the ems_clusters element would be selected to get the name ?

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+    //vms/name
+    private String vmName;
```
jonathanvila: which one of the vms element would be selected to get the name ?

jonathanvila:
I assume the structure is VM centric, therefore we would have 1 object of that class per VM in the structure of the payload.
Am I right ?

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
m-brophy: I'll change this to be 'systemServicesNames',  a collection of Strings containing all the different 'name' values, which is the only field actually required by the rules.

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
m-brophy: I'll represent each file by a key, value pair consisting of the name and contents elements

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
m-brophy: To keep the data sent to a minimum that's required we've agreed to keep this as it is.

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
jonathanvila: //vms/ems_ref

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
m-brophy: the spreadsheet describes how to find the correct ems_cluster

m-brophy:
> I assume the structure is VM centric, therefore we would have 1 object of that class per VM in the structure of the payload.
> Am I right ?

That's correct

src/main/java/org/jboss/xavier/analytics/pojo/input/VMWorkloadInventoryModel.java
```diff
@@ -0,0 +1,179 @@
...
+    //hardware/guest_os_full_name
+    private String guestOSFullName;
+    //vms/has_rdm_disk
+    private boolean hasRdmDisk;
+    //count of nics object within the vms/hardware
+    private Integer nicsCount;
+    //hardware/disks/filename
+    private Collection<String> vmDiskFilenames;
+    private Collection<String> systemServicesNames;
+    private Collection<Map<String,String>> files;
```
mrizzi: I feel like having just a Map (without the Collection) would be enough to accomplish what we decided during the review meeting.
Each file will be an entry with `key = file name` and `content = file content`.
In this way we should be able to collect all the files from the payload in a Map (without the "wrapping" Collection)

```suggestion
    private Map<String,String> files;
```

mrizzi (commented):
There's just one further change to be applied and then it will be good.

mrizzi (approved):
thanks for taking care of creating the bean @m-brophy.
As agreed i'm going to approve & merge this PR even if the build doesn't work due to maven.repository.redhat.com outage

PR #18, opened 2019-07-24
MIGENG-131 WorkloadInventoryReportModel bean

mrizzi:
https://jira.coreos.com/browse/MIGENG-131

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
@@ -1,0 +1,215 @@
+package org.jboss.xavier.analytics.pojo.output;
+
+import com.fasterxml.jackson.annotation.JsonBackReference;
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+import javax.persistence.ElementCollection;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
```diff
@@ -0,0 +1,215 @@
...
+    private String cluster;
+    private String vmName;
+    private String osName;
+    private String osDescription;
+    private BigDecimal diskSpace;
+    private Integer memory;
+    private Integer cpuCores;
+    @ElementCollection(fetch = FetchType.EAGER)
+    private Set<String> workloads;
+    private String complexity;
```
carlosthe19916, line 55: @mrizzi Do you think it is worth changing the type of `complexity` field to an `Enum` instead of `String`? I think the possible values of `complexity` is limited.

After this question is answered I think I will merge your PR. I think that travis CI is failing because of external problems

carlosthe19916: Answer: Once we will be more stable we should make the complexity to become an enum but for now let's keep it as it is defined.

carlosthe19916 (approved):
As agreed I'm going to approve & merge this PR even if the build doesn't work due to maven.repository.redhat.com outage

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
mrizzi: Thanks @carlosthe19916 for reporting this and i can confirm once we will be sure about the `complexity` values and meaning, we will move to have it better defined using an enumeration.

PR #16, opened 2019-07-19
Hotfix deployment 18july

jonathanvila:

PR #19, opened 2019-07-24
MIGENG-103 Reporting API - Workload Inventory Report

mrizzi:
https://jira.coreos.com/browse/MIGENG-103

WIP because without tests

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
+++ b/src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
@@ -1,0 +1,15 @@
+package org.jboss.xavier.integrations.jpa.repository;
+
+import org.jboss.xavier.analytics.pojo.output.InitialSavingsEstimationReportModel;
+import org.jboss.xavier.analytics.pojo.output.WorkloadInventoryReportModel;
+import org.jboss.xavier.integrations.jpa.projection.InitialSavingsEstimationReportSummary;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface WorkloadInventoryReportRepository extends JpaRepository<WorkloadInventoryReportModel, Long>
+{
+    Page<WorkloadInventoryReportModel> findByAnalysisId(Long analysisId, Pageable pageable);
+}
--- a/src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
+++ b/src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
@@ -1,8 +1,6 @@
 package org.jboss.xavier.integrations.jpa.repository;
 
-import org.jboss.xavier.analytics.pojo.output.InitialSavingsEstimationReportModel;
 import org.jboss.xavier.analytics.pojo.output.WorkloadInventoryReportModel;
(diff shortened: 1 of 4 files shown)


mrizzi:
With commit 2ce0a62
* added tests (with and without `page` and `size` parameters) for the new `/report/{id}/workload-inventory` endpoint
* added the missing tests for `/report/{id}` and `/report/{id}/initial-saving-estimation` as well
* following https://github.com/project-xavier/xavier-integration/pull/16/files#diff-508a34a6f039810be3427e43274e4ed4R51, switched the `<bean>` components from using `beanType` attribute to `ref` in order to have the tests to work
* fixed the tests' method names `mainRouteBuilder*` to be `xmlRouteBuilder*` to get consistency with the naming convention

src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
```diff
@@ -0,0 +1,15 @@
+package org.jboss.xavier.integrations.jpa.repository;
+
+import org.jboss.xavier.analytics.pojo.output.InitialSavingsEstimationReportModel;
```
carlosthe19916: This import is unnecessary

src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
```diff
@@ -0,0 +1,15 @@
...
+import org.jboss.xavier.analytics.pojo.output.WorkloadInventoryReportModel;
+import org.jboss.xavier.integrations.jpa.projection.InitialSavingsEstimationReportSummary;
```
carlosthe19916: This import is unnecessary

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
```diff
@@ -80,7 +84,87 @@ public void mainRouteBuilder_RestReport_NotSummaryParamGiven_ShouldCallFindRepor
...
+    public void xmlRouteBuilder_RestReportId_IdParamGiven_ShouldCallFindReportSummaryById() throws Exception {
+        //Given
+        camelContext.setTracing(true);
+        camelContext.setAutoStartup(false);
+
+        //When
+        camelContext.start();
+        camelContext.startRoute("report-get-details");
+        Map<String, Long> variables = new HashMap<>();
+        Long one = new Long(1);
```
carlosthe19916: I would suggest changing this to `Long one = 1L;` but it is Ok if you just ignore this. There are a couple of lines with the same definition in this class.

carlosthe19916 (commented):
In general, this PR looks great to me. It would be great to merge this PR.

src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
mrizzi: thanks, i'll remove it

src/main/java/org/jboss/xavier/integrations/jpa/repository/WorkloadInventoryReportRepository.java
mrizzi: thanks, i'll remove it

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
mrizzi: I'll change this :+1:

carlosthe19916 (approved):
This PR looks great to me. I tested it using the UI and postman and it worked well!

PR #20, opened 2019-07-24
MIGENG-365 - End2end integration test

jonathanvila:
https://issues.redhat.com/browse/MIGENG-365

This E2E test is based in these assumptions : 
1. creating a real Kafka broker
2. creating a real ActiveMQ Artemis broker
3. creating a real Ingress instance
4. creating a real KIE instance

To do this it's used : 
* TestContainers : to create containers for Kafka, AMQ Artemis, Ingress, PostgreSQL, Minio, Localstack, KIE [https://www.testcontainers.org]


pom.xml
```diff
@@ -62,9 +69,14 @@
...
         </exclusions>
         </dependency>
+
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-undertow</artifactId>
+        </dependency>    
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-activemq</artifactId>
```
mrizzi: did you find an issue in the current code that fails for not having that dependency?

pom.xml
```diff
@@ -38,11 +38,18 @@
...
+
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.10.19</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mock-server</groupId>
+            <artifactId>mockserver-netty</artifactId>
```
mrizzi: shouldn't this have `<scope>test</scope>`?

src/main/java/org/jboss/xavier/integrations/route/filter/HeaderFilter.java
```diff
@@ -9,7 +9,7 @@
 
     @Override
     public boolean applyFilterToCamelHeaders(String arg0, Object arg1, Exchange arg2) {
-        return false;
+        return true;
```
mrizzi: It seems a radical change from `false` to `true`: does this solve an issue?

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -91,8 +92,9 @@ public void configure() {
                 .to("log:INFO?showBody=true&showHeaders=true")
                 .end();
 
-        from("kafka:" + kafkaHost + "?topic={{insights.kafka.upload.topic}}&brokers=" + kafkaHost + "&autoOffsetReset=latest&autoCommitEnable=true")
+        from("kafka:" + kafkaHost + "?topic={{insights.kafka.upload.topic}}&brokers=" + kafkaHost + "&autoOffsetReset=earliest&autoCommitEnable=true")
```
mrizzi, line 124: can we have a test to test this change is consistent?

mrizzi (commented):
I added some comments for some doubts and questions i have about some changes.
I found 2 issues with `EndToEndIT` test:

1. the `EndToEndIT` test is not executed during the Maven build because it's missing the conventional `Test` prefix or suffix in the file name 
2. I force the execution locally and it went into an infinite loop with a SELinux alert and these exceptions 
```
11:37:59.019 [testcontainers-ryuk] WARN  o.t.utility.ResourceReaper - Can not connect to Ryuk at localhost:32769
java.net.ConnectException: Connection refused (Connection refused)
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at java.net.Socket.<init>(Socket.java:434)
	at java.net.Socket.<init>(Socket.java:211)
	at org.testcontainers.utility.ResourceReaper.lambda$start$1(ResourceReaper.java:113)
	at java.lang.Thread.run(Thread.java:748)
11:37:59.019 [testcontainers-ryuk] WARN  o.t.utility.ResourceReaper - Can not connect to Ryuk at localhost:32769
java.net.ConnectException: Connection refused (Connection refused)
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at java.net.Socket.<init>(Socket.java:434)
	at java.net.Socket.<init>(Socket.java:211)
	at org.testcontainers.utility.ResourceReaper.lambda$start$1(ResourceReaper.java:113)
	at java.lang.Thread.run(Thread.java:748)
11:37:59.019 [testcontainers-ryuk] WARN  o.t.utility.ResourceReaper - Can not connect to Ryuk at localhost:32769
java.net.ConnectException: Connection refused (Connection refused)
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at java.net.Socket.<init>(Socket.java:434)
	at java.net.Socket.<init>(Socket.java:211)
	at org.testcontainers.utility.ResourceReaper.lambda$start$1(ResourceReaper.java:113)
	at java.lang.Thread.run(Thread.java:748)
```
Maybe i need some more configuration locally to run it?

src/test/java/org/jboss/xavier/integrations/EndToEndIT.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    @NotNull
+    private HttpEntity<MultiValueMap<String, Object>> getRequestEntityForRESTCall() throws IOException {
+
+        // Headers
+        HttpHeaders headers = new HttpHeaders();
+        headers.setContentType(MediaType.MULTIPART_FORM_DATA);
+        headers.set("x-rh-insights-request-id", "2544925e825b4f3f9418c88556541776");
+        headers.set("x-rh-identity", "eyJlbnRpdGxlbWVudHMiOnsiaW5zaWdodHMiOnsiaXNfZW50aXRsZWQiOnRydWV9LCJvcGVuc2hpZnQiOnsiaXNfZW50aXRsZWQiOnRydWV9LCJzbWFydF9tYW5hZ2VtZW50Ijp7ImlzX2VudGl0bGVkIjpmYWxzZX0sImh5YnJpZF9jbG91ZCI6eyJpc19lbnRpdGxlZCI6dHJ1ZX19LCJpZGVudGl0eSI6eyJpbnRlcm5hbCI6eyJhdXRoX3RpbWUiOjAsImF1dGhfdHlwZSI6Imp3dC1hdXRoIiwib3JnX2lkIjoiNjM0MDA1NiJ9LCJhY2NvdW50X251bWJlciI6IjE0NjAyOTAiLCJ1c2VyIjp7ImZpcnN0X25hbWUiOiJNYXJjbyIsImlzX2FjdGl2ZSI6dHJ1ZSwiaXNfaW50ZXJuYWwiOnRydWUsImxhc3RfbmFtZSI6IlJpenppIiwibG9jYWxlIjoiZW5fVVMiLCJpc19vcmdfYWRtaW4iOmZhbHNlLCJ1c2VybmFtZSI6Im1yaXp6aUByZWRoYXQuY29tIiwiZW1haWwiOiJtcml6emkrcWFAcmVkaGF0LmNvbSJ9LCJ0eXBlIjoiVXNlciJ9fQ==");
```
m-brophy: Just a query -  is there a reason we're using different identity here to the one in src/test/resources/platform.upload.xavier-with-targz.json?

src/main/java/org/jboss/xavier/integrations/route/filter/HeaderFilter.java
m-brophy: This method seems to be saying we want to apply filters to Headers but in 
src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java configure() we've made a change to say 

> .to("http4://oldhost?headerFilterStrategy=#noFilter")

and the two strategies don't seem to match

m-brophy (commented):
I've left a couple of queries about things I don't understand

jonathanvila:
1. You are right about the Test suffix. In fact this is an IT test ( that's why it has de IT suffix ) and should be excluded from the regular tests and only included in the failsafe plugin. 
--> What do you prefer : execute this test separatedly from the regular tests once every time, or do it every time we execute the tests.

jonathanvila:
2. In order to execute correctly the test, I needed to have an updated version of the docker installation. I needed to : 
*  sudo dnf remove docker
*  sudo dnf install docker-ce

And then remove images locally in order to avoid having duplicates : 
* sudo docker rmi $(docker images -q) --force
* sudo service docker stop
* sudo rm -rf /var/lib/docker
* sudo service docker start

In order to avoid the --insecure error : 
* sudo vi  /etc/docker/daemon.json
* { "insecure-registries": ["172.30.0.0/16"] }
* Then restart Docker && sudo oc cluster up  ( not using a previous base folder )
https://github.com/openshift/origin/issues/8997

pom.xml
jonathanvila: No, probably it's a temporary change while fighting with spring boot.
I will check if it's really needed.
:+1:

pom.xml
jonathanvila: Totally :+1:

src/main/java/org/jboss/xavier/integrations/route/filter/HeaderFilter.java
jonathanvila: Let me check :+1:

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
jonathanvila: Sure, it will take time though.
You mean having a test comparing the 2 approaches, because we didnt know if the previous one was doing what we wanted....to do that we need several messages, and a retarded connection by the kafka listener in order to see which one is consuming more messages

src/test/java/org/jboss/xavier/integrations/EndToEndIT.java
jonathanvila: Good catch.
Let me check why :+1:

src/main/java/org/jboss/xavier/integrations/route/filter/HeaderFilter.java
jonathanvila: Good catch. In fact the #noFilter is the name of the bean. So to be consistent with the name we should not do any filtering.
Let me check :+1:

mrizzi:
>     1. You are right about the Test suffix. In fact this is an IT test ( that's why it has de IT suffix ) and should be excluded from the regular tests and only included in the failsafe plugin.
>        --> What do you prefer : execute this test separatedly from the regular tests once every time, or do it every time we execute the tests.

For sure in this phase i see the most benefit coming from having it to run every time we open a PR to test every change.

mrizzi:
>     1. In order to execute correctly the test, I needed to have an updated version of the docker installation. I needed to :
> 
> 
>     * sudo dnf remove docker
> 
>     * sudo dnf install docker-ce
>       And then remove images locally in order to avoid having duplicates :
> 
>     * sudo docker rmi $(docker images -q) --force
> 
>     * sudo service docker stop
> 
>     * sudo rm -rf /var/lib/docker
> 
>     * sudo service docker start

I tried `sudo dnf install docker-ce` but it seems like this package is coming from a repo i haven't got.
```
[mrizzi@localhost xavier-integration](jonathanvila-end2end-integration-test)$ sudo dnf install docker-ce
Last metadata expiration check: 0:03:20 ago on Thu 25 Jul 2019 02:21:48 PM CEST.
No match for argument: docker-ce
Error: Unable to find a match
[mrizzi@localhost xavier-integration](jonathanvila-end2end-integration-test)$ dnf list docker*
Last metadata expiration check: 0:10:44 ago on Thu 25 Jul 2019 02:20:01 PM CEST.
Installed Packages
docker.x86_64                                                  2:1.13.1-67.git1185cfd.fc30                           @updates
docker-common.x86_64                                           2:1.13.1-67.git1185cfd.fc30                           @updates
docker-compose.noarch                                          1.22.0-4.fc30                                         @updates
docker-rhel-push-plugin.x86_64                                 2:1.13.1-67.git1185cfd.fc30                           @updates
Available Packages
docker-anaconda-addon.x86_64                                   0.4-10.fc30                                           fedora  
docker-client-java.noarch                                      8.11.7-4.fc30                                         fedora  
docker-devel.noarch                                            2:1.13.1-67.git1185cfd.fc30                           updates 
docker-distribution.x86_64                                     2.6.2-9.git48294d9.fc30                               fedora  
docker-fish-completion.x86_64                                  2:1.13.1-67.git1185cfd.fc30                           updates 
docker-latest.x86_64                                           2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-devel.noarch                                     2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-fish-completion.x86_64                           2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-logrotate.x86_64                                 2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-rhsubscription.x86_64                            2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-unit-test.x86_64                                 2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-vim.x86_64                                       2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-latest-zsh-completion.x86_64                            2:1.13.1-42.git1185cfd.fc30                           fedora  
docker-logrotate.x86_64                                        2:1.13.1-67.git1185cfd.fc30                           updates 
docker-lvm-plugin.x86_64                                       2:1.13.1-67.git1185cfd.fc30                           updates 
docker-novolume-plugin.x86_64                                  2:1.13.1-67.git1185cfd.fc30                           updates 
docker-vim.x86_64                                              2:1.13.1-67.git1185cfd.fc30                           updates 
docker-zsh-completion.x86_64                                   2:1.13.1-67.git1185cfd.fc30                           updates 
[mrizzi@localhost xavier-integration](jonathanvila-end2end-integration-test)$ 
```
Are you sure this will solve the issue?
I'm a little bit concerned of moving away from Fedora's `docker` to switch to `docker-ce` so i would like to be sure this will solve.

jonathanvila:
> Are you sure this will solve the issue?
> I'm a little bit concerned of moving away from Fedora's `docker` to switch to `docker-ce` so i would like to be sure this will solve.

This is where my docker-ce comes from : 

> Name         : docker-ce
> Epoch        : 3
> Version      : 19.03.0
> Release      : 3.fc30
> Architecture : x86_64
> Size         : 103 M
> Source       : docker-ce-19.03.0-3.fc30.src.rpm
> Repository   : @System
> From repo    : docker-ce-stable
> Summary      : The open-source application container engine
> URL          : https://www.docker.com

jonathanvila:
> > ```
> > 1. You are right about the Test suffix. In fact this is an IT test ( that's why it has de IT suffix ) and should be excluded from the regular tests and only included in the failsafe plugin.
> >    --> What do you prefer : execute this test separatedly from the regular tests once every time, or do it every time we execute the tests.
> > ```
> 
> For sure in this phase i see the most benefit coming from having it to run every time we open a PR to test every change.

Done

mrizzi:
@jonathanvila i feel like we could kill two birds with one stone.
I took some time to investigate this following your hint about a property to solve the issue i experienced.
It turned out:

- the `docker` version i have (`1.13.1`) is supported since the minimum is `1.10` (see [testcontainers - General Docker requirements](https://www.testcontainers.org/supported_docker_environment/))
- the issue about Ryuk has been reported in https://github.com/testcontainers/testcontainers-java/issues/892
- the issue has been solved in https://github.com/testcontainers/testcontainers-java/issues/700#issuecomment-457541087 with the `TESTCONTAINERS_RYUK_DISABLED` env prop
- i tested this locally and everything worked for him

So i sent you https://github.com/jonathanvila/xavier-integration/pull/1 on your branch so that you can evaluate if this new property works with `docker-ce` as well because, if it works, we will know we can work with both Fedora's `docker` and `docker-ce`.

Hi @bsideup, would you mind adding your thoughts on this situation?
Using Fedora's `docker` (version `1.13.1`, with minimum required being `v1.10`) we experience an infinite loop of `[testcontainers-ryuk] WARN  o.t.utility.ResourceReaper - Can not connect to Ryuk at localhost:32769
java.net.ConnectException: Connection refused (Connection refused)` exceptions.
I don't want to look scared of or against changing from Fedora provided `docker` to `docker-ce` but i would really like to understand where the issue is and why, even if the version is supported by testContainers, Fedora's docker doesn't work so that we can take an informed decision.

Thanks in advance for your help.

src/main/java/org/jboss/xavier/integrations/route/filter/HeaderFilter.java
jonathanvila: Done

src/test/java/org/jboss/xavier/integrations/EndToEndIT.java
jonathanvila: Made all the same to be consistent
Done

src/main/java/org/jboss/xavier/integrations/route/filter/HeaderFilter.java
jonathanvila: Fixed and moved back to false as it was, and tests work.
Done

pom.xml
jonathanvila: Done

bsideup:
@mrizzi the Ryuk issue in https://github.com/testcontainers/testcontainers-java/issues/892 has nothing to do with yours, I think. "Cannot connect to Ryuk" may happen due to many various reasons, and, in case of BitBucket, it is due their security model.

`TESTCONTAINERS_RYUK_DISABLED` was introduced for CI environments with strong security requirements *and* an ability to perform cleanups after the job automatically (public CIs)

We may need to update our docs in the minimum supported Docker version.
Since 1.13.0 is 2.5 years old, and being rapidly changed, it may mean we have no other choice.

I would suggest trying running Ryuk as described in https://github.com/testcontainers/moby-ryuk to verify that Ryuk is the problem first:
```bash
docker run -v /var/run/docker.sock:/var/run/docker.sock -p 8080:8080 quay.io/testcontainers/ryuk:0.2.3
nc localhost 8080 # may need to use the gateway's IP instead of "localhost"
```
(make sure you run it from the same environment as the tests', especially if you do it inside a container)

mrizzi:
@bsideup thanks for the info.
I tried to run it locally with this result:
```
[mrizzi@localhost ~]$ docker run -v /var/run/docker.sock:/var/run/docker.sock -p 8080:8080 quay.io/testcontainers/ryuk:0.2.3
2019/07/26 14:34:53 Starting on port 8080...
panic: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/_ping: dial unix /var/run/docker.sock: connect: permission denied

goroutine 1 [running]:
main.main()
	/go/src/github.com/testcontainers/moby-ryuk/main.go:31 +0xe2e
```
I checked the groups and that the `hello-world` container runs to exclude any kind of missing/wrong configuration in my environment:
```
[mrizzi@localhost ~]$ groups 
mrizzi wheel libvirt docker
[mrizzi@localhost ~]$ docker run hello-world

Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
    (amd64)
 3. The Docker daemon created a new container from that image which runs the
    executable that produces the output you are currently reading.
 4. The Docker daemon streamed that output to the Docker client, which sent it
    to your terminal.

To try something more ambitious, you can run an Ubuntu container with:
 $ docker run -it ubuntu bash

Share images, automate workflows, and more with a free Docker ID:
 https://hub.docker.com/

For more examples and ideas, visit:
 https://docs.docker.com/get-started/
```
In the end, reading [Dan Walsh's post](https://danwalsh.livejournal.com/74095.html), i tried again running it with the `--privileged` option and it started without any exception and letting me executing the commands in https://github.com/testcontainers/moby-ryuk#usage:
```
[mrizzi@localhost ~]$ docker run --privileged -v /var/run/docker.sock:/var/run/docker.sock -p 8080:8080 quay.io/testcontainers/ryuk:0.2.3
2019/07/26 14:51:31 Starting on port 8080...
2019/07/26 14:51:40 Connected
2019/07/26 14:51:48 Adding {"health":{"unhealthy":true},"label":{"testing=true":true}}
2019/07/26 14:51:53 Adding {"label":{"something":true}}
2019/07/26 14:52:01 EOF
2019/07/26 14:52:01 Disconnected
2019/07/26 14:52:08 Connected
2019/07/26 14:52:08 Adding {"label":{"something_else":true}}
2019/07/26 14:52:08 EOF
2019/07/26 14:52:08 Disconnected
2019/07/26 14:52:18 Timed out waiting for connection
2019/07/26 14:52:18 Deleting {"health":{"unhealthy":true},"label":{"testing=true":true}}
2019/07/26 14:52:19 Deleting {"label":{"something":true}}
2019/07/26 14:52:19 Deleting {"label":{"something_else":true}}
2019/07/26 14:52:19 Removed 0 container(s), 3 network(s), 130 volume(s)
```
So, since Ryuk has to run as a privileged container in this case, i tried to add the `ryuk.container.privileged=true` property in my local `~/.testcontainers.properties` file (following https://www.testcontainers.org/features/configuration/#customizing-ryuk-resource-reaper) and everything worked fine.
This seems to me a better solution than `TESTCONTAINERS_RYUK_DISABLED` property.
@bsideup wdyt?

bsideup:
@mrizzi yeah, definitely! Nice catch :)
`ryuk.container.privileged=true` was added for the environments like yours, where the socket requires special permissions.

Just make sure it is set in home folder's config file, not classpath's. **We plan to deprecate the ability of setting environment-specific properties in classpath files**

mrizzi:
Thanks for the quick feedback @bsideup and for the help in figuring out this issue.
We will add a note in our REAME, something like:
```
If you're running the tests using Fedora's docker, please add the ryuk.container.privileged=true property in ~/.testcontainers.properties file in your home folder.
```
In this way we have the developer to get the freedom to use both `docker` and `docker-ce` :wink: 
Two birds killed with one stone :smile: 
Thanks again for the help @bsideup

pom.xml
jonathanvila: Removed

jonathanvila:
@mrizzi @bsideup how this approach affects our deployment to Travis CI ? AFAIK Travis uses docker-ce version, so apparently no need to change anything there, but well, someone in the team probably should use docker-ce (if everything works fine) in order to have the same version/branch of docker as the CI runner.
wdyt ?

bsideup:
@jonathanvila Travis does not need this flag. We do our best to work with any supported Docker version and keep the experience consistent (security makes it harder sometimes, that "privileged" flag is a good example), so I don't think you **need** someone to run the same Docker version as Travis does

mrizzi:
@jonathanvila i agree with you: no need to change anything for Travis CI.
AFAIK, you already have `docker-ce` locally so we are safe for the moment.
In case something wrong happens in Travis CI, we know that to reproduce it we will have to investigate with `docker-ce` (and hence install it locally if we are missing it).

What has to be done for sure in this PR is adding the description of the issue, the exception and how to fix it in the REDAME file so that the knowledge is shared outside of this PR.
Could you update the README please?

src/main/resources/application-test.properties
```diff
@@ -11,4 +11,16 @@ spring.datasource.continue-on-error=true
 insights.properties=dummy
 
 spring.datasource.url = jdbc:h2:mem:test
-spring.jpa.properties.hibernate.dialect = org.hibernate.dialect.H2Dialect
\ No newline at end of file
+spring.jpa.properties.hibernate.dialect = org.hibernate.dialect.H2Dialect
```
mrizzi, line 14: Having the power of testcontainer available, i think it would be better to replace H2 database with PostgreSQL 9.6 so that the test will be using the same database used in the running environment.

pom.xml
```diff
@@ -221,6 +228,18 @@
...
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.12.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>kafka</artifactId>
+            <version>1.11.3</version>
```
mrizzi: Version [`1.12.0`](https://repo1.maven.org/maven2/org/testcontainers/kafka/1.12.0/) has been released so we can keep the `org.testcontainers`'s artifacts version aligned to `1.12.0`

mrizzi (commented):
Some further improvement to get the test as close as possible to the running environment using the same DB.

pom.xml
bsideup: FYI there is also `org.testcontainers:testcontainers-bom`

src/main/resources/application-test.properties
bsideup: Especially given that it is as easy as:
```properties
spring.datasource.url=jdbc:tc:postgresql:9.6.8:///databasename
spring.datasource.driver-class-name=org.testcontainers.jdbc.ContainerDatabaseDriver
```

:)

mrizzi: thanks for the support @bsideup, really appreciated :+1:

jonathanvila:
@mrizzi README file updated.

README.md
```diff
@@ -107,5 +107,12 @@ Snippets of calls :
 * See status of encription on a bucket : `aws s3api get-bucket-encryption --bucket xavier-dev`
 * Enable encription on a bucket : `aws s3api put-bucket-encryption --bucket xavier-dev --server-side-encryption-configuration '{"Rules": [{"ApplyServerSideEncryptionByDefault": {"SSEAlgorithm": "AES256"}}]}' `  
 
+
+# Testing
+For the End2End test a docker container test framework is used : [TestContainers](www.testcontainers.org).
```
bsideup: FYI ;)

```suggestion
For the End2End test a docker container test framework is used : [Testcontainers](www.testcontainers.org).  
```

jonathanvila: +1

jonathanvila:
Take into account that ScanRunModel has been modified in order to make tests pass : 
`public Boolean getType() {
        return type;

//        String typeString = "Virt Platform";
//        return typeString + (type?" + SmartState": "");   //TODO move back this before pushing
    }

    public void setType(Boolean type) {
        this.type = type;
    }

    public void setType(String strType) {
        if (isBoolean(strType)) {
            setType(Boolean.parseBoolean(strType));
        } else {
            setType(strType.toLowerCase().contains("smartstate"));
        }
    }`

src/main/resources/application-test.properties
jonathanvila: For the unit tests we will still use H2 and for the End2End we will use PostgreSQL

carlosthe19916:
@jonathanvila In `EndToEndTest.java` I've tried to copy and paste the method `public void end2endTest()` so we have 2 tests (2methods) inside the same class. When I execute the tests using the IDE or using the command `mvn test -Dtest=EndToEndTest` the first method always succeeds and the second one fails.
- Both methods contain the same code.
- See the screenshot for better understanding

Screenshot from 2019-12-13 15-28-29

Is there something I'm doing wrong? I was expecting to see the 2 tests to pass successfully.

carlosthe19916:
@jonathanvila Could you please add to the README file which ports the e2e tests are using? Just as documentation, so we all can stop other services that are running in those ports

jonathanvila:
> @jonathanvila In `EndToEndTest.java` I've tried to copy and paste the method `public void end2endTest()` so we have 2 tests (2methods) inside the same class. When I execute the tests using the IDE or using the command `mvn test -Dtest=EndToEndTest` the first method always succeeds and the second one fails.
> 
> * Both methods contain the same code.
> * See the screenshot for better understanding
> 
> Screenshot from 2019-12-13 15-28-29
> 
> Is there something I'm doing wrong? I was expecting to see the 2 tests to pass successfully.

Well, first thing is that the test has not been created to have multiple test methods.
But, let me check locally.

jonathanvila:
> @jonathanvila Could you please add to the README file which ports the e2e tests are using? Just as documentation, so we all can stop other services that are running in those ports

Testcontainers uses Random ports.... what's the point of documenting which are the poirts used ?

carlosthe19916:
> > @jonathanvila Could you please add to the README file which ports the e2e tests are using? Just as documentation, so we all can stop other services that are running in those ports
> 
> Testcontainers uses Random ports.... what's the point of documenting which are the poirts used ?

@jonathanvila I see, so Testcontainer is using random ports. I was suggesting that because It happened to me that I had a couple of other services running with docker and when I ran the tests of `EndToEndTest.java` it threw an exception. I guess it was just a coincidence.

carlosthe19916:
@jonathanvila Adding some comments:
1. This is interesting: I've edited the `.travis.yml` and removed the code to install a custom version of docker and docker-compose; in short words, I've ran the tests using the default Docker and docker-compose versions of Travis, and all the tests passed successfully. You can see it here https://travis-ci.org/carlosthe19916/xavier-integration/jobs/625693280/config
2. After executing the tests in `EndToEndTest.java` there are test files remaining like: `src/test/resources/insights-ingress-go/*` and `src/test/resources/insights-ingress-go/ingressRepo.zip`; is it possible to delete those files after the test or create them in the `target` folder? this is just to avoid committing those files accidentally.

.github/workflows/maven.yml
```diff
@@ -0,0 +1,17 @@
+name: Java CI
```
carlosthe19916: What is the purpose of this file? I think it is not used.

pom.xml
```diff
@@ -379,4 +423,94 @@
...
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+            <releases>
+                <enabled>true</enabled>
+                <updatePolicy>never</updatePolicy>
+            </releases>
+        </pluginRepository>
+
+    </pluginRepositories>
```
carlosthe19916: I've removed this block of code from pom.xml and it worked without any problem in Travis. Here is the Travis build https://travis-ci.org/carlosthe19916/xavier-integration/jobs/625773720 and here is the GitHub commit https://github.com/carlosthe19916/xavier-integration/commit/3c9433adc2498f3df73c78fe2c88ace33c009d25 ...  I wonder if we can remove this block? What do you think?

.github/workflows/maven.yml
jonathanvila: everything under .github is for GitHub Actions . I already suggested to move away from Travis and use Gihub as the unique tool. It's true that change is not directly related to the PR.

pom.xml
jonathanvila: It was working before in Travis with that block. What's the issue with adding the repositories ? 
If there is no artifact that comes from those repositories then I agree we dont need it.

jonathanvila: In fact, these are the repositories we are using, but we are using a customized settings.xml file for travis..... I think we should move away from having things in the settings.xml and put it declaratively in our pom

jonathanvila: Also, for the github Actions we need to have these repositories , either following the same approach as travis or CI agnostic putting them in the pom
It's true that this change is related to the Github migration task and not to the End2End

carlosthe19916:
@jonathanvila Thanks for the last commit, I can see that the last build in Travis failed with the same error I've found locally: `Condition with lambda expression in org.jobss.xavier.integrations.EndToEndTest was not fulfilled within 10000 milliseconds`

This is the Travis Log:
[log.txt](https://github.com/project-xavier/xavier-integration/files/3982854/log.txt)

Here is a screenshot from my local machine.
Screenshot from 2019-12-19 10-28-41

I've tested locally and it seems the error appears when my Laptop is slow (When I don't have all the CPU resources available). I guess it is because the containers are taking too much time for starting and it causes the problem.

I wonder if increasing the time from `10000` will solve the problem or not, what do you think?

carlosthe19916:
@jonathanvila I've tried to deploy your PR in CI and the pod is not able to start because `Could not resolve placeholder 'minio.host' in value "${minio.host}"` error, could you please make the necessary changes to `application.properties`, `application-test.properties`, or any other file you require to make it deployable in CI?

Here is the log of CI: 
[analytics-integration-276-hgvrd.log](https://github.com/project-xavier/xavier-integration/files/3982896/analytics-integration-276-hgvrd.log)

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ScanRunModel.java
```diff
@@ -100,12 +101,25 @@ public void setDate(Date date) {
         this.date = date;
     }
 
-    public String getType() {
-        String typeString = "Virt Platform";
-        return typeString + (type?" + SmartState": "");
+    public Boolean getType() {
+        return type;
```
carlosthe19916, line 108: As we agreed on, this change will require to create a Ticket for changing the UI so there are no visual changes. For now, I can confirm that this block of code will change the `Scans run` table of the `Workload Migration summary` report... I'm writing just not to forget to create the ticket

Screenshot from 2019-12-19 10-35-17

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -121,20 +124,26 @@ public void configure() throws Exception {
...
         from("direct:download-file")
                 .routeId("download-file")
-                .setHeader("Exchange.HTTP_URI", simple("${body.url}"))
+                .setHeader("tempHTTP_URI", simple("${body.url}"))
+                .process(e -> {
+                    String headerURI = "tempHTTP_URI";
+                    e.getIn().setHeader(headerURI, e.getIn().getHeader(headerURI, String.class).replace("minio:9000", minio_host));
+                })
+                .setHeader("Host", constant("minio:9000"))
+                .setHeader("Exchange.HTTP_URI", simple("${header.tempHTTP_URI}"))
```
carlosthe19916: I think we are mixing Test code with the original code. What about something like the code below, so we are not mixing but just adding new code.
```
from("direct:download-file")
                .routeId("download-file")
                .setHeader("Exchange.HTTP_URI", simple("${body.url}"))
// Here we can write the code for the testing environment
                .choice()
                    .when(exchange -> ((String) exchange.getIn().getBody()).contains("minio:9000"))
                        .process(e -> e.getIn().setHeader("Exchange.HTTP_URI",  minio_host))
                        .setHeader("Host", constant("minio:9000"))
                .end()

                .convertBodyTo(FilePersistedNotification.class)
```
I'm proposing this just to be sure which code will run in Test environment

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -70,6 +70,9 @@
 
     private List<Integer> httpSuccessCodes = Arrays.asList(HttpStatus.SC_OK, HttpStatus.SC_CREATED, HttpStatus.SC_ACCEPTED, HttpStatus.SC_NO_CONTENT);
 
+    @Value("${minio.host:x}")
+    private String minio_host;
```
carlosthe19916: Maybe a comment to remember that is is just for testing environment

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
jonathanvila: Yep, I'm already changing that ... in another way, removing that piece of code from the final route.
Thank you

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
jonathanvila: I have moved it to the test class.

src/main/java/org/jboss/xavier/integrations/SpringConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+@Configuration
+@Profile("!test")
```
carlosthe19916, line 13: When `test` profile is activated then `SpringConfiguration.java` is valid and also `TestConfigurationS3.java`, in theory, both of them should be valid, but I'm not sure if it will create an error or not... What about using `@Profile("default")` instead of `@Profile("!test")` what do you think?

jonathanvila: SpringConfiguration and TestConfigurationS3 are creating the AWS_S3 bean.
The test one needs to inject the URL of the host.
SpringConfiguration will only be enbled if profile is different than TEST ( note the ! in the profile name ).

src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_S3Test.java
```diff
@@ -28,14 +28,14 @@
     @Test
     @Ignore
     public void s3Test() throws Exception {
```
carlosthe19916, line 30: I know this test is ignored even in the master branch but if this Test Class is used somehow I would suggest to fix it and remove the @Ignore; otherwise, we can delete this in order to clean the source code. What do you think?

.travis.yml
```diff
@@ -8,13 +8,40 @@ addons:
...
+  - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+  - sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/1.22.0/docker-compose-Linux-x86_64 > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+  - docker-compose --version
+  - docker info
```
carlosthe19916, line 31: I would suggest using the default versions of Travis so when Travis will upgrade the Docker and Docker Compose versions we will have it without any change in our side

src/test/java/org/jboss/xavier/integrations/route/WorkloadSummaryReportRoutes_DirectCalculateVMWorkloadInventoryModelTest.java
```diff
@@ -372,11 +372,9 @@ public void DirectCalculateVMWorkloadInventoryModel_ShouldPersistScanRunModel()
         Assert.assertNotNull(scanRunModels);
         Assert.assertEquals(2, scanRunModels.size());
 
-        scanRunModels.stream().filter(model -> model.getId() % 2 == 0).forEach(srm ->
-                Assert.assertEquals("Virt Platform", srm.getType()));
+        Assert.assertTrue(scanRunModels.stream().filter(model -> model.getId() % 2 == 0).allMatch(srm -> !srm.getType()));
```
carlosthe19916, line 375: This is just a reminder to not forget to create the ticket for changing the UI.

src/test/resources/drools-kie-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: '2'
```
carlosthe19916: I might be wrong but it seems to me that this file is not used in the end2end tests. We can remove this file, What do you think?

src/test/resources/end2end_test_docker_curl_commands.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
```
carlosthe19916: I might be wrong but it seems to me that this file is not used in the end2end tests. We can remove this file, What do you think?

carlosthe19916 (commented):
A couple of more comments

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
```diff
@@ -0,0 +1,524 @@
...
+        private static String getContainerHost(GenericContainer container) {
+            return container.getContainerIpAddress() + ":" + container.getFirstMappedPort();
+        }
+    }
+
+    @Inject
+    CamelContext camelContext;
+
+    @Inject
+    JmsTemplate jmsTemplate;
```
carlosthe19916: We can remove this since it is not used anywhere

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
```diff
@@ -0,0 +1,524 @@
...
+    @Value("${test.timeout.performance:60000}") // 1 minute
+    private int timeoutMilliseconds_PerformaceTest;
+
+    @Value("${test.timetout.ics:10000}") // 10 seconds
+    private int timeoutMilliseconds_InitialCostSavingsReport;
+
+    @Value("${minio.host:x}")
+    private String minio_host;
+
+    public static class Initializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
```
carlosthe19916, line 162: Since we have the limitation of writing just one Test method by Class, then I would suggest moving the `Initializer` class outside the `EndToEndTest.java` class since it might be used by future Test Class. What do you think?

carlosthe19916: As we discussed; my idea is to make some parts of the Class reusable by future End2End test classes. However, I agree that for now there are no other End2End tests in mind so we can postpone the refactoring of this class

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ScanRunModel.java
mrizzi: @jonathanvila and @carlosthe19916 sorry, just for my benefit.
The old `getType` method's implementation has changed: it returns a `Boolean` and not the `String` the UI was showing (highlighted in the above screenshot).
Is there, in this PR, a new method that return the same `String` the old `getType` method provided?

src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_S3Test.java
jonathanvila: Well this test is here because it was handy to test the S3 component. We can remove it , but I wanted to leave it there in order to have the code if we need to test more things with the component.

.travis.yml
jonathanvila: Well, I wanted to fix a version that we know it works.... yes using the default we will have updates, for the good and for the bad, as we dont know if Testcontainers will cope with the same version of the docker installed in Travis.

src/test/resources/drools-kie-compose.yml
jonathanvila: This file is there because it helped to start the compose with KIE and test the commands. We can remove it , but I left it here in order to have the yaml if we need to do further tests with the compose.

src/test/resources/end2end_test_docker_curl_commands.sh
jonathanvila: The same as with the kie compose. It was helpful to have it in order to recreate the infrastracture with the compose. We can remove it , but I thought it was interesting to have it.

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
jonathanvila: +1

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
jonathanvila: As we talked, moving the initializer out of the class will convert it in a non static class .... and also it needs the values from the containers started ( ip and port ) there fore we need to have the initializer and the test class on sync. So I dont see much benefit on moving the class out and it will complicate things a bit.
Also, I think reusable is interesting but only when you need to reuse....

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ScanRunModel.java
jonathanvila: No , there is no method to return the string needed. The UI ticket is only going to show the text that comes from the report or it will convert the boolean to the String it was being sent ? When we said the UI was going to change because of this change I assumed it was because the UI would be in charge of that change. If it's the backend who has to send the string in the report , then that method is missing at the moment.

mrizzi: Discussed a little bit more about this with @carlosthe19916 because to me having a `boolean type` field was not clear what it referred to.
The previous `String type` was providing the `Type` label value to the shown in the UI (screenshot above) but now a boolean value is hard to understand what it refers to just calling it `type`.
I think it could be improved having the field to be renamed to `smartStateEnabled` so that it's clear the boolean refers to Smart State begin enabled in the analysis.

Then we will change the UI to manage this boolean field accordingly (https://issues.redhat.com/browse/MIGENG-370)

carlosthe19916: @mrizzi @jonathanvila I've created a PR (on Jonathan's branch) for renaming `type` to `smartStateEnabled`; however, this change is just using `@JsonProperty("smartStateEnabled")` so it won't affect the current database model. It is not trivial to change the database model since hibernate won't update the database automatically.  for us. Maybe in the future, we can introduce a database version control manager like [flyway](https://flywaydb.org/)

jonathanvila: Before it was also a boolean on the mutator setType , and it was not clear at all too, then we had a String accessor getType not returning the internal value but transforming it using a hard coded string. So we lived with unclearity also before. 

My suggestion is to solve the type of the attribute now and in a separate ticket rename the attribute, and execute the rename into the database or implement Flyway (or any other DB change tool ) in order to move forward with this PR

carlosthe19916 (approved):
Thanks @jonathanvila for applying all the changes we discussed during the last couple of weeks :)... as far as I'm concern, this PR looks good to me so I'll approve it.

src/test/resources/minio-bucket-creation-commands.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+/usr/bin/mc config host add myminio http://$1 BQA2GEXO711FVBVXDWKM uvgz3LCwWM3e400cDkQIH/y1Y4xgU4iV91CwFSPC;
+/usr/bin/mc mb myminio/insights-upload-perma;
+/usr/bin/mc mb myminio/insights-upload-rejected;
+/usr/bin/mc policy download myminio/insights-upload-perma;
```
mrizzi: it seem that locally this command didn't work for me since in the log it prints the usage output:
```
16:31:12.000 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT: Added `myminio` successfully.
16:31:12.000 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT: Bucket created successfully `myminio/insights-upload-perma`.
16:31:12.000 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT: Bucket created successfully `myminio/insights-upload-rejected`.
16:31:12.000 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT: Name:
16:31:12.000 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT:   mc policy - manage anonymous access to buckets and objects
16:31:12.076 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT: USAGE:
16:31:12.077 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT:   mc policy set [FLAGS] PERMISSION TARGET
16:31:12.077 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT:   mc policy set-json [FLAGS] FILE TARGET
16:31:12.077 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT:   mc policy get [FLAGS] TARGET
16:31:12.077 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT:   mc policy get-json [FLAGS] TARGET
16:31:12.077 [tc-okhttp-stream-2142129482] INFO  o.j.xavier.integrations.EndToEndTest - [MINIO-MC-LOG] STDOUT:   mc policy list [FLAGS] TARGET
```

mrizzi (commented):
@jonathanvila i ran it locally 3 times and i found one issue with a command in a script.

mrizzi (approved):
Latest changes worked for me.
Thanks @jonathanvila for this improvement to help us improve code's stability and maintainability.
Thanks @bsideup for the help and support :+1:

PR #22, opened 2019-07-29
MIGENG-157: MA - Reporting - Analysis (Report) List Delete

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-157
added DELETE /report/{id} endpoint:
- When a report does not exist, then it will return 404 status code.
- When a report exists, then it will delete the report.

### DIFF
--- a/src/main/resources/spring/camel-context.xml
+++ b/src/main/resources/spring/camel-context.xml
@@ -67,6 +67,9 @@
                         <when>
                             <simple>${body} != null</simple>
                             <bean ref="analysisService" method="deleteById(${header.id})" />
+                            <setHeader headerName="Exchange.HTTP_RESPONSE_CODE">
+                                <constant>204</constant>
+                            </setHeader>
                         </when>
                         <otherwise>
                             <setHeader headerName="Exchange.HTTP_RESPONSE_CODE">
--- a/src/main/resources/spring/camel-context.xml
+++ b/src/main/resources/spring/camel-context.xml
@@ -68,12 +68,12 @@
                             <simple>${body} != null</simple>
                             <bean ref="analysisService" method="deleteById(${header.id})" />
                             <setHeader headerName="Exchange.HTTP_RESPONSE_CODE">
-                                <constant>204</constant>
+                                <simple resultType="String">${type:javax.servlet.http.HttpServletResponse.SC_NO_CONTENT}</simple>
                             </setHeader>
                         </when>
                         <otherwise>
                             <setHeader headerName="Exchange.HTTP_RESPONSE_CODE">
-                                <constant>404</constant>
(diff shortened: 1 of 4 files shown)


src/main/resources/spring/camel-context.xml
```diff
@@ -59,6 +59,26 @@
...
                 </route>
             </get>
+            <delete uri="/{id}">
+                <description>Delete a report</description>
+                <route id="report-delete">
+                    <bean ref="analysisService" method="findById(${header.id})"/>
+                    <choice>
+                        <when>
+                            <simple>${body} != null</simple>
+                            <bean ref="analysisService" method="deleteById(${header.id})" />
```
mrizzi, line 69: As agreed we could change the approach to return no content and `204` HTTP code (`No Content`) in case the delete request worked successfully.
In this way we are also compliant with [RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7)

mrizzi (commented):
Added a comment after we talked about the response format.

src/main/resources/spring/camel-context.xml
carlosthe19916: @mrizzi my last commit should make the endpoint return 204 HTTP code with an empty body. Thanks for noticing this

src/main/resources/spring/camel-context.xml
```diff
@@ -59,6 +59,29 @@
...
+                            <setHeader headerName="Exchange.HTTP_RESPONSE_CODE">
+                                <constant>204</constant>
```
mrizzi: Could you replace this with [`HttpServletResponse.SC_NO_CONTENT`](https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html#SC_NO_CONTENT)?

src/main/resources/spring/camel-context.xml
```diff
@@ -59,6 +59,29 @@
...
+                            </setHeader>
+                        </when>
+                        <otherwise>
+                            <setHeader headerName="Exchange.HTTP_RESPONSE_CODE">
+                                <constant>404</constant>
```
mrizzi: Could you replace this with [`HttpServletResponse.SC_NOT_FOUND`](https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html#SC_NOT_FOUND)

carlosthe19916: @mrizzi I got your point and you're right it's better to use constants. My last commit changed it

mrizzi (approved):
The code looks good and everything works locally.
Thanks @carlosthe19916 :+1:

PR #23, opened 2019-07-29
MIGENG-148: MA - Reporting - Analysis (Report) List Filtering

carlosthe19916:
add filterText QueryParam to `/report` to allow filtering reports/analysis. I took the chance of removing `summary` query param too.

https://jira.coreos.com/browse/MIGENG-148


mrizzi:
Waiting for https://jira.coreos.com/browse/MIGENG-155 to be implemented before reviewing this PR since we need a real workflow to test this.

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
```diff
@@ -63,38 +63,62 @@ public void setup() {
...
         camelContext.start();
         camelContext.startRoute("reports-get-all");
-        Map<String, String> variables = new HashMap<>();
-        variables.put("summary", "true");
-        restTemplate.getForEntity(camel_context + "report?summary={summary}", String.class, variables);
+        restTemplate.getForEntity(camel_context + "report", String.class);
 
         //Then
-        verify(analysisService).findReports(0, 10);
+        verify(analysisService).findReports(anyInt(), anyInt());
```
mrizzi: I think we should check that default values (`page="0"` and `size="10"`) for the params are applied in this test

mrizzi (commented):
It works fine: i've just added a comment for a check in one test.

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
carlosthe19916: My last commit changed anyInt() by static pagination values

mrizzi (approved):
Thanks @carlosthe19916: it looks good to me :+1:

PR #24, opened 2019-07-29
MIGENG-164 Added indexes to InitialSavingsEstimationReportModel and WorkloadInventoryReportModel

mrizzi:
https://jira.coreos.com/browse/MIGENG-164

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/InitialSavingsEstimationReportModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/InitialSavingsEstimationReportModel.java
@@ -11,9 +11,11 @@ import javax.persistence.Entity;
 import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
+import javax.persistence.Index;
 import javax.persistence.JoinColumn;
 import javax.persistence.MapsId;
 import javax.persistence.OneToOne;
+import javax.persistence.Table;
 import java.util.Date;
 
 /**
@@ -21,11 +23,19 @@ import java.util.Date;
  */
 
 @Entity
+@Table(
+        indexes = {
+                @Index(name = "InitialSavingsEstimationReportModel_" +
+                        InitialSavingsEstimationReportModel.ANALYSIS_ID + "_index",
+                        columnList = InitialSavingsEstimationReportModel.ANALYSIS_ID, unique = false)
+        }
+)
 public class InitialSavingsEstimationReportModel
         implements
         java.io.Serializable {
 
     static final long serialVersionUID = 1L;
+    static final String ANALYSIS_ID = "analysis_id";
 
     @Id
     @GeneratedValue(strategy = javax.persistence.GenerationType.AUTO, generator = "INITIALSAVINGSESTIMATIONREPORTMODEL_ID_GENERATOR")
@@ -39,7 +49,7 @@ public class InitialSavingsEstimationReportModel
     private Long id;
 
     @OneToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "analysis_id")
+    @JoinColumn(name = ANALYSIS_ID)
     @JsonBackReference
     private AnalysisModel analysis;
 
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
@@ -4,40 +4,49 @@ import com.fasterxml.jackson.annotation.JsonBackReference;
 import org.hibernate.annotations.GenericGenerator;
 import org.hibernate.annotations.Parameter;
 
+import javax.persistence.CollectionTable;
 import javax.persistence.ElementCollection;
 import javax.persistence.Entity;
 import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/output/InitialSavingsEstimationReportModel.java
```diff
@@ -11,21 +11,31 @@
...
 /**
  * This class was automatically generated by the data modeler tool.
  */
 
 @Entity
+@Table(
+        indexes = {
+                @Index(name = "InitialSavingsEstimationReportModel_" +
+                        InitialSavingsEstimationReportModel.ANALYSIS_ID + "_index",
+                        columnList = InitialSavingsEstimationReportModel.ANALYSIS_ID, unique = false)
```
carlosthe19916, line 30: `unique` value is by default `false` so no need to write it again.

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
```diff
@@ -4,40 +4,49 @@
...
-import java.util.List;
 import java.util.Set;
 
 @Entity
 @Transactional
+@Table(
+        indexes = {
+                @Index(name = "WorkloadInventoryReportModel_" +
+                        WorkloadInventoryReportModel.ANALYSIS_ID + "_index",
+                        columnList = WorkloadInventoryReportModel.ANALYSIS_ID, unique = false)
```
carlosthe19916, line 29: `unique` value is by default `false` so no need to write it again.

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
```diff
@@ -51,13 +60,28 @@
...
+    @CollectionTable(
+            indexes = {
+                    @Index(columnList = "workload_inventory_report_model_id", unique = false)
+            }
+    )
     private Set<String> recommendedTargetsIMS;
     @ElementCollection(fetch = FetchType.EAGER)
+    @CollectionTable(
+            indexes = {
+                    @Index(columnList = "workload_inventory_report_model_id", unique = false)
```
carlosthe19916, line 82: `unique` value is by default `false` so no need to write it again.

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
```diff
@@ -51,13 +60,28 @@
...
+            }
+    )
     private Set<String> workloads;
     private String complexity;
     // with "IMS" suffix in case one day we will have
     // their "AMM" counterparts
     @ElementCollection(fetch = FetchType.EAGER)
+    @CollectionTable(
+            indexes = {
+                    @Index(columnList = "workload_inventory_report_model_id", unique = false)
```
carlosthe19916, line 75: `unique` value is by default `false` so no need to write it again.

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
```diff
@@ -51,13 +60,28 @@
     private Integer memory;
     private Integer cpuCores;
     @ElementCollection(fetch = FetchType.EAGER)
+    @CollectionTable(
+            indexes = {
+                    @Index(columnList = "workload_inventory_report_model_id", unique = false)
```
carlosthe19916, line 65: `unique` value is by default `false` so no need to write it again.

carlosthe19916 (changes requested):
@mrizzi  Overall this PR looks great to me; your changes are improving the performance of the DELETE operation over AnalysisModels!

There are just some optional changes I would like to do (see the comments).

src/main/java/org/jboss/xavier/analytics/pojo/output/InitialSavingsEstimationReportModel.java
mrizzi: TBH i think it's better to set the value we want without relying on default value in this case in order to avoid any issue if and when the default value will change in the future.

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
mrizzi: As above

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
mrizzi: As above

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
mrizzi: As above

src/main/java/org/jboss/xavier/analytics/pojo/output/WorkloadInventoryReportModel.java
mrizzi: As above

carlosthe19916 (approved):
@mrizzi OK I got your point about using `unique = false` even though it is set by default: **We should write it in order to avoid future changes of Hibernate where this value could change**

PR #27, opened 2019-07-31
MIGENG-153: Column Sorting - Workload Inventory Report

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-153
add sorting to /report/{id}/workload-inventory

Because there are a lot of parameters (with filtering the number of parameters will increase) which will be passed to `workloadInventoryReportService` I decided to create two additional camel routes which will take headers, create java classes, and finally add new headers fo the Exchange Object.

The idea is to wrap some headers into classes like: `SortBean`, `PageBean`, `FiltersBean` and then pass these objects to our `Service` methods.


src/test/java/org/jboss/xavier/integrations/route/ToBeanRouter_DirectToPaginationBean.java
```diff
@@ -0,0 +1,162 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(CamelSpringBootRunner.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+@MockEndpointsAndSkip("")
+@UseAdviceWith // Disables automatic start of Camel context
+@SpringBootTest(classes = {Application.class})
+@ActiveProfiles("test")
+public class ToBeanRouter_DirectToPaginationBean {
```
mrizzi: Please add the `Test` suffix otherwise the test won't be executed
```suggestion
public class ToBeanRouter_DirectToPaginationBeanTest {
```

src/test/java/org/jboss/xavier/integrations/route/ToBeanRouter_DirectToSortBean.java
```diff
@@ -0,0 +1,161 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(CamelSpringBootRunner.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+@MockEndpointsAndSkip("")
+@UseAdviceWith // Disables automatic start of Camel context
+@SpringBootTest(classes = {Application.class})
+@ActiveProfiles("test")
+public class ToBeanRouter_DirectToSortBean {
```
mrizzi: Please add the `Test` suffix otherwise the test won't be executed
```suggestion
public class ToBeanRouter_DirectToSortBeanTest {
```

mrizzi (changes requested):
The changes look code: the indexes for speeding up the queries have been added and i think the `Bean` approach to manage the query parameters makes sense.
There's something to be fixed: the test classes need to have `Test` as suffix and/or prefix to be executed during the standard Mvane build. I suggested to add the suffix for consistency with other tests in the project.

carlosthe19916:
> The changes look code: the indexes for speeding up the queries have been added and i think the `Bean` approach to manage the query parameters makes sense.
> There's something to be fixed: the test classes need to have `Test` as suffix and/or prefix to be executed during the standard Mvane build. I suggested to add the suffix for consistency with other tests in the project.

@mrizzi Thanks for reviewing this PR, I've already applied your suggestions and now the Test Classes have `Test` suffix.

mrizzi (approved):
Thanks @carlosthe19916: the tests i've done locally using the UI from https://github.com/project-xavier/xavier-ui/pull/32 are now fine and the build works :+1:

PR #30, opened 2019-08-04
MIGENG-155 Create Analysis Model and Set Reports objects

jonathanvila:
One important change here : 
    * in order to be able to add elements to the workloadinventoryreports collection , there are 2 options, or manage the hibernate session by hand, or move the fetch type to eager. Also as we are always going to work with the all reports, I dont see why we benefit from having lazy fetching on this.

https://jira.coreos.com/browse/MIGENG-155


mrizzi (commented):
Thanks for the work done @jonathanvila.
I created #33 to apply the changes required after testing in CI env

PR #33, opened 2019-08-06
MIGENG-155 Create Analysis Model and Set Reports objects

mrizzi:
Follows up #30 

* added the mandatory (in the overall process) `analysisId` value to the `x-rh-identity` header sent to the Insights Upload service
* fixed `CamelFileName` input field reference
* added `AbstractInputModel` bean to propagate the `analysisId` field over the JMS queues (with `xstream.processAnnotations(AbstractInputModel.class);` to avoid having it in XML request to DM)
* changed `MA_metadata` casting to be `(Map<String,String>)` instead of `(Map<String,Object>)` since this is the type of `Map` set as header in https://github.com/project-xavier/xavier-integration/blob/master/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java#L104
* introduced `MainRouteBuilder.ANALYSIS_ID` static field to use everywhere the same value
* removed `MA_metadata` headers from JMS test because headers are not sent to JMS so `MA_metadata` won't be available in JMS listener
* added `inserted` and `lastUpdate` fields to `AnalysisModel`
* changed the `/report` REST endpoint to retrieve the analysis list (updating the tests)


mrizzi:
I'm going to merge this PR even if i open it since i did it just to add a fix to the code developed in #30  for which i'm the reviewer.

PR #34, opened 2019-08-09
MIGENG-184 WS Report skeleton routes and tests

mrizzi:
### Issue
https://jira.coreos.com/browse/MIGENG-184
### Changes
- added the first version of the `WorkloadSummaryReportModel` with the `SummaryModel`
- the `SummaryModel` has `NamedNativeQuery` and `SqlResultSetMapping` to provide an example about how to create new bean for the boxes in the WSR with each one having its own query to be created
- the `SummaryRepository` has the declaration of the method to execute  the `SummaryModel.calculateSummaryModels` native query (the name follows a Spring convention)
- enhanced the `VMWorkloadInventoryRoutes.calculate-vmworkloadinventory` route so that once it finishes successfully (`.onCompletion().onCompleteOnly()`), it starts the `WorkloadSummaryReportRoutes.aggregate-vmworkloadinventory` route
- `VMWorkloadInventoryRoutes.calculate-vmworkloadinventory` route: i don't like it too much because i'm sure there's a better way to do it with Camel using, maybe, [JPA poller feauters](https://camel.apache.org/jpa.html) and [aggregator EIP](https://camel.apache.org/aggregator.html).  
Anyway it works and it can be controlled with 2 properties: 
   - `report.workload.summary.polling.delay` defines the time in milliseconds between each poll (default `5000`)
   - `report.workload.summary.polling.max-attempts` defines the max number of attempts before throwing an exception to avoid having a thread stuck in the pool loop (default `120`)  
   - So the poller waits for 5 minutes for all the expected `WorkloadInventoryReportModel` to be persisted in the DB
   - `WorkloadSummaryReportRoutes_DirectAggregateVMWorkloadInventoryModelTest` test persists `WorkloadInventoryReportModel` in the DB one after the other so the poller is tested 
   - `WorkloadSummaryReportRoutes_DirectAggregateVMWorkloadInventoryModelTest_ThrowException` test checks that the poller throws exactly the expected exception so that we're safe that there's no infinite poller loop
- added the `/report/{id}/workload-summary` REST endpoint
- removed some sample rules and method not used any more to avoid any further confusion

### Workflow to create a new bean for a box in the WSR
In order to create the code for the `Foo` box in the WSR, you need to:
1. copy&paste `SummaryModel.java` with name `FooModel.java`
1. add the needed fields to the bean
1. change the `@NamedNativeQuery` fields
1. change the `@SqlResultSetMapping` mapping
1. add the `FooModel` to the `WorkloadSummaryReportModel`
1. copy&paste the `SummaryRepository` with name `FooRepository`
1. add the method to the `FooRepository` that matches the `@NamedNativeQuery` annotation
1. copy&paste `SummaryService.java` with name `FooService.java`
1. add the method created early in the `FooRepository`
1. enhance the code in the `aggregate-vmworkloadinventory` route to execute the method and so the native query
1. enhance the `WorkloadSummaryReportRoutes_DirectCalculateVMWorkloadInventoryModelTest` to check the new `FooModel` has been persisted in the DB with the expected values

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/AnalysisModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/AnalysisModel.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.hibernate.annotations.GenericGenerator;
 import org.hibernate.annotations.Parameter;
 import org.jboss.xavier.analytics.pojo.output.workload.inventory.WorkloadInventoryReportModel;
+import org.jboss.xavier.analytics.pojo.output.workload.summary.WorkloadSummaryReportModel;
 
 import javax.persistence.CascadeType;
 import javax.persistence.Entity;
@@ -37,6 +38,10 @@ public class AnalysisModel
     @OneToMany(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
     @JsonIgnore
     private List<WorkloadInventoryReportModel> workloadInventoryReportModels;
+
+    @OneToOne(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
+    @JsonIgnore
+    private WorkloadSummaryReportModel workloadSummaryReportModels;
 
     private String reportName;
     private String reportDescription;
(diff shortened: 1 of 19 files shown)


mrizzi:
As agreed, the review has been done during the meeting we had earlier today and the changes to the code and to the PR's description ("Workflow to create a new bean for a box in the WSR") cover the changes requested during the review.

PR #35, opened 2019-08-16
MIGENG-185: Business Rules Implementation - WSR Complexity

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-185
Added:
- `ComplexityModel`
- `ComplexityService`
- `ComplexityRepository`

Enhanced:

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ComplexityModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ComplexityModel.java
@@ -1,0 +1,120 @@
+package org.jboss.xavier.analytics.pojo.output.workload.summary;
+
+import com.fasterxml.jackson.annotation.JsonBackReference;
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+import javax.persistence.*;
+
+@SqlResultSetMapping(
+        name = "mappingComplexityModels",
+        classes = @ConstructorResult(
+                targetClass = ComplexityModel.class,
+                columns = {
+                        @ColumnResult(name = "easy", type = Integer.class),
+                        @ColumnResult(name = "medium", type = Integer.class),
+                        @ColumnResult(name = "difficult", type = Integer.class),
+                        @ColumnResult(name = "unknown", type = Integer.class)
+                }
+        )
+)
+
+@NamedNativeQuery(
+        name = "ComplexityModel.calculateComplexityModels",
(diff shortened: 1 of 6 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ComplexityModel.java
```diff
@@ -0,0 +1,120 @@
...
+                        @ColumnResult(name = "medium", type = Integer.class),
+                        @ColumnResult(name = "difficult", type = Integer.class),
+                        @ColumnResult(name = "unknown", type = Integer.class)
+                }
+        )
+)
+
+@NamedNativeQuery(
+        name = "ComplexityModel.calculateComplexityModels",
+        query = "select sum(case when lower(complexity)='easy' then 1 else 0 end) as easy, sum(case when lower(complexity)='medium' then 1 else 0 end) as medium, sum(case when lower(complexity)='difficult' then 1 else 0 end) as difficult, sum(case when (complexity is null or lower(complexity)='unknown') then 1 else 0 end) as \"unknown\" from workload_inventory_report_model where analysis_id = :analysisId",
```
mrizzi:
```suggestion
        query = "select sum(case when lower(complexity)='easy' then 1 else 0 end) as easy, sum(case when lower(complexity)='medium' then 1 else 0 end) as medium, sum(case when lower(complexity)='hard' then 1 else 0 end) as hard, sum(case when (complexity is null or lower(complexity)='unknown') then 1 else 0 end) as \"unknown\" from workload_inventory_report_model where analysis_id = :analysisId",
```

mrizzi (commented):
The overall code looks good (also the test) :+1: 
There's one change needed to be consistent with the required values for the complexity: instead of `difficult` we are required to use `hard`

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ComplexityModel.java
carlosthe19916: Thanks for discovering this @mrizzi , I've already changed "difficult" by "hard" not just in the query but in the Class field.

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/ComplexityModel.java
```diff
@@ -0,0 +1,120 @@
...
+    public void setUnknown(Integer unknown) {
+        this.unknown = unknown;
+    }
+
+    @Override
+    public String toString() {
+        return "SummaryModel{" +
+                "id=" + id +
+                ", report=" + report +
+                ", easy='" + easy + '\'' +
```
mrizzi:
```suggestion
                ", easy='" + easy +
```

mrizzi (commented):
Just a little change to clean the code.

PR #36, opened 2019-08-16
MIGENG-186: Business Rules Implementation - WSR Recommended Targets

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-186
Added:
- `RecommendedTargetsIMSModel`
- `RecommendedTargetsService`
- `RecommendedTargetsRepository`

Enhanced:

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
@@ -21,7 +21,7 @@ import javax.persistence.*;
 
 @NamedNativeQuery(
         name = "RecommendedTargetsIMSModel.calculateRecommendedTargetsIMS",
-        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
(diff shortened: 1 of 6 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
```diff
@@ -0,0 +1,121 @@
...
+                        @ColumnResult(name = "rhv", type = Integer.class),
+                        @ColumnResult(name = "osp", type = Integer.class),
+                        @ColumnResult(name = "rhel", type = Integer.class)
+                }
+        )
+)
+
+@NamedNativeQuery(
+        name = "RecommendedTargetsIMSModel.calculateRecommendedTargetsIMS",
+        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
```
mrizzi: `rhel` has to be changed to `convert2rhel` and i think it's fine to change it just in the query since the UI will show `RHEL` label and not `Convert2RHEL` so i think it's worth only changing the query and nothing else in the code.
```suggestion
        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='convert2rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
```

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
```diff
@@ -0,0 +1,121 @@
...
+    public void setOsp(Integer osp) {
+        this.osp = osp;
+    }
+
+    @Override
+    public String toString() {
+        return "SummaryModel{" +
+                "id=" + id +
+                ", report=" + report +
+                ", rhv='" + rhv + '\'' +
```
mrizzi:
```suggestion
                ", rhv='" + rhv +
```

mrizzi (commented):
I suggested a change in the query to have it working with the values inserted but the rules.

carlosthe19916:
@mrizzi I've just committed your suggested changes to the query. I hope this PR is ok after done it :)

mrizzi (approved):
Thanks for this enhancement @carlosthe19916 :+1:

PR #39, opened 2019-08-20
MIGENG-188: Business Rules Implementation - WSR FLags

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-188
Added:
- `FlagModel`
- `FlagService`
- `FlagRepository`

Enhanced:

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/FlagModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/FlagModel.java
@@ -1,0 +1,125 @@
+package org.jboss.xavier.analytics.pojo.output.workload.summary;
+
+import com.fasterxml.jackson.annotation.JsonBackReference;
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
+
+import javax.persistence.*;
+
+@SqlResultSetMapping(
+        name = "mappingFlagModels",
+        classes = @ConstructorResult(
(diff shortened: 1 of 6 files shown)


mrizzi:
As decided during our meeting, the `Flags` and `Assessment` values should be provided from the REST Endpoint using a lookup table in the DB to translate the raw flags coming from the rules into the values to be used in the UI.

Example:

1. in the DB there's the value `RDM` in the `flagsims` in the `workload_inventory_report_model_flagsims` table
2. the native query `FlagModel.calculateFlagModels` will persist the `RDM` value in the `flag_model` table with `RHEL` as `os_name`
3. when the UI will call the `/report/{id}/workload-summary/flags` endpoint (from https://github.com/project-xavier/xavier-integration/pull/42) the REST endpoint will decode the `RDM`+`RHEL` pair using a new lookup table in the DB getting the values for the `Flags` and `Assessment` columns for the UI that, in this example, will be `Raw Device Mapping` and `Manual data migration required`.

So the lookup table in the DB has to have 3 columns:
1. flag IMS value (e.g. `RDM`)
2. OS name (e.g. `RHEL`)
2. flag description (e.g.  `Raw Device Mapping`)
3. assessment description (e.g. `Manual data migration required`)

mrizzi (approved):
As agreed, this PR is fine as it is right now so thanks a lot @carlosthe19916 for the work done.

The lookup table design & implementation will be done in another issue and PR so that we can carefully evaluate the best solution.

PR #40, opened 2019-08-20
Migeng 201 fix tar gz capability

jonathanvila:
Replaced the convertBodyTo(String.class) with convertBodyTo(byte[].class)
Fixed test not passing due to past changes
Removed Content-Type check as it can be binary/octet-stream sometimes
https://jira.coreos.com/browse/MIGENG-201


mrizzi (approved):
Thanks @jonathanvila for fixing this :+1: 
I tested it with both with `cloudforms-export-v1.tar.gz` and `cloudforms-export-v1.json` (for backward compatibility) and everything looks good to me :wink:

mrizzi (commented):
This also fixes https://jira.coreos.com/browse/MIGENG-129

PR #43, opened 2019-08-23
Migeng 156 cover exceptions and mark analysis as failed

jonathanvila:
Setting the Analysis as FAILED if these cases occur :
1. exception or http response <> 200 on Uploading to Insights Service
2. exception on analysing the file or http response <> 200 on downloading the file from Insights Service
3. exception on creating the report or http response <> 200 calling KIE server

https://jira.coreos.com/browse/MIGENG-156

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
+++ b/src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
@@ -17,6 +17,12 @@ import java.util.Date;
 @Component
 public class AnalysisService
 {
+    public enum STATUS {
+        FAIL,
+        IN_PROGRESS,
+        CREATED
+    }
+
     @Autowired
     AnalysisRepository analysisRepository;
 
@@ -37,7 +43,7 @@ public class AnalysisService
         analysisModel.setReportName(reportName);
         analysisModel.setInserted(new Date());
         analysisModel.setLastUpdate(new Date());
-        analysisModel.setStatus("IN_PROGRESS");
+        analysisModel.setStatus(STATUS.IN_PROGRESS.toString());
 
         return analysisRepository.save(analysisModel);
     }
@@ -59,7 +65,7 @@ public class AnalysisService
         analysisModel.setWorkloadSummaryReportModels(reportModel);
         reportModel.setAnalysis(analysisModel);
         // TODO remove this since it's just a temporary workaround to change the status
-        analysisModel.setStatus("CREATED");
+        analysisModel.setStatus(STATUS.CREATED.toString());
         analysisRepository.save(analysisModel);
     }
 
@@ -75,4 +81,9 @@ public class AnalysisService
         return analysisRepository.findByReportNameIgnoreCaseContaining(filterText.trim(), pageable);
     }
 
+    public void updateStatus(String status, Long id) {
+        AnalysisModel analysisModel = findById(id);
+        analysisModel.setStatus(status);
+        analysisRepository.save(analysisModel);
+    }
 }
--- a/src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
+++ b/src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
@@ -18,7 +18,7 @@ import java.util.Date;
 public class AnalysisService
 {
     public enum STATUS {
-        FAIL,
+        FAILED,
         IN_PROGRESS,
         CREATED
     }
--- a/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
+++ b/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
@@ -107,8 +107,7 @@ public class MainRouteBuilder extends RouteBuilder {
                     .to("http4://" + uploadHost + "/api/ingress/v1/upload")
                     .choice()
                         .when(not(isResponseSuccess()))
-                            .to("direct:mark-analysis-fail")
-                            .end()
+                            .throwException(org.apache.commons.httpclient.HttpException.class, "Unsuccessful response from Insights Upload Service")
                 .endDoTry()
                 .doCatch(Exception.class)
                     .to("log:error?showCaughtException=true&showStackTrace=true")
@@ -117,7 +116,7 @@ public class MainRouteBuilder extends RouteBuilder {
 
         from("direct:mark-analysis-fail").id("markAnalysisFail")
                 .process(e -> analysisService.updateStatus(AnalysisService.STATUS.FAILED.toString(), Long.parseLong((String) e.getIn().getHeader("MA_metadata", Map.class).get(ANALYSIS_ID))))
-                .end();
+                .stop();
 
 
         from("kafka:" + kafkaHost + "?topic={{insights.kafka.upload.topic}}&brokers=" + kafkaHost + "&autoOffsetReset=latest&autoCommitEnable=true")
@@ -140,7 +139,7 @@ public class MainRouteBuilder extends RouteBuilder {
                             .to("direct:unzip-file")
                             .log("File ${header.CamelFileName} success")
                         .otherwise()
-                            .to("direct:mark-analysis-fail")
+                            .throwException(org.apache.commons.httpclient.HttpException.class, "Unsuccessful response from Insights Download Service")
                 .endDoTry()
                 .doCatch(Exception.class)
                     .to("log:error?showCaughtException=true&showStackTrace=true")
--- a/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
+++ b/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
@@ -33,6 +33,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.apache.camel.builder.PredicateBuilder.not;
+
 /**
  * A Camel Java8 DSL Router
  */
@@ -104,8 +106,9 @@ public class MainRouteBuilder extends RouteBuilder {
                     .removeHeaders("Camel*")
                     .to("http4://" + uploadHost + "/api/ingress/v1/upload")
                     .choice()
-                        .when(header(Exchange.HTTP_RESPONSE_CODE).isNotEqualTo(HttpStatus.SC_OK))
-                            .to("direct:mark-analysis-fail").end()
+                        .when(not(isResponseSuccess()))
+                            .to("direct:mark-analysis-fail")
+                            .end()
                 .endDoTry()
                 .doCatch(Exception.class)
                     .to("log:error?showCaughtException=true&showStackTrace=true")
@@ -113,7 +116,7 @@ public class MainRouteBuilder extends RouteBuilder {
                 .end();
 
         from("direct:mark-analysis-fail").id("markAnalysisFail")
-                .process(e -> analysisService.updateStatus(AnalysisService.STATUS.FAIL.toString(), (Long) e.getIn().getHeader("MA_metadata", Map.class).get(ANALYSIS_ID)))
+                .process(e -> analysisService.updateStatus(AnalysisService.STATUS.FAILED.toString(), Long.parseLong((String) e.getIn().getHeader("MA_metadata", Map.class).get(ANALYSIS_ID))))
                 .end();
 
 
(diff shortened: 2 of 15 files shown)


src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
```diff
@@ -17,6 +17,12 @@
 @Component
 public class AnalysisService
 {
+    public enum STATUS {
```
carlosthe19916, line 20: - I'd suggest to move the STATUS enum to his own file and locate it into the package `org.jboss.xavier.analytics.pojo.output`
- I suggest also to change the AnalysisModel#status to: `@Enumerated(EnumType.STRING) private String status`. In this way, we will avoid using `STATUS.IN_PROGRESS.toString()` and we can use directly the ENUM value

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -96,14 +99,26 @@ public void configure() {
...
+                        .when(not(isResponseSuccess()))
+                            .throwException(org.apache.commons.httpclient.HttpException.class, "Unsuccessful response from Insights Upload Service")
+                .endDoTry()
+                .doCatch(Exception.class)
+                    .to("log:error?showCaughtException=true&showStackTrace=true")
+                    .to("direct:mark-analysis-fail")
                 .end();
 
+        from("direct:mark-analysis-fail").id("markAnalysisFail")
+                .process(e -> analysisService.updateStatus(AnalysisService.STATUS.FAILED.toString(), Long.parseLong((String) e.getIn().getHeader("MA_metadata", Map.class).get(ANALYSIS_ID))))
```
mrizzi, line 120: I'm wondering if such a general purpose method could be add as `DefaultErrorHandler` as well so that, for whatever other use case not covered in this PR, we know we have it managed implementing something based on `DefaultErrorHandler`'s [example](https://camel.apache.org/manual/latest/defaulterrorhandler.html#DefaultErrorHandler-Example) like (consider it like pseudo-code):
```java
onException(Exception.class)
        .handled(true)
        .process(e -> analysisService.updateStatus(AnalysisService.STATUS.FAILED.toString(), Long.parseLong((String) e.getIn().getHeader("MA_metadata", Map.class).get(ANALYSIS_ID))));
```

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -148,23 +174,23 @@ public void configure() {
...
-                    .to("log:error?showCaughtException=true&showStackTrace=true")
-                    .setBody(simple("Exception on parsing Cloudforms file"))
+                .transform().method("calculator", "calculate(${body}, ${header.MA_metadata})")
+                .log("Message to send to AMQ : ${body}")
+                .to("jms:queue:uploadFormInputDataModel")
                 .end();
     }
 
+    private Predicate isResponseSuccess() {
+        return header(Exchange.HTTP_RESPONSE_CODE).isEqualTo(HttpStatus.SC_OK);
```
mrizzi: I suppose this implementation covers the remote services we call right now and maybe, to avoid changes in the future, it's worth adding other `2XX` codes as well in this method.
For example i feel like it's worth adding `201` (`SC_CREATED`), `202` (`SC_ACCEPTED`) and `204` (`SC_NO_CONTENT`).

src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
mrizzi, line 20: Thanks for adding this `enum`, it was needed :+1:

mrizzi (commented):
I'm still testing it but looking at the code, i reported 3 comments/suggestions.

carlosthe19916 (changes requested):
I think we can change the field `status` in the `AnalysisModel` class in order to use ENUMs instead of Strings. @jonathanvila What do you think?

carlosthe19916:
@jonathanvila I've just found a problem and I'd like you to verify it, please.
When a valid `.json` is uploaded, the first status of the analysis is `FAILED` and after a couple of seconds it becomes `CREATED`. I think the first status should be `IN_PROGRESS` and then `CREATED` or `FAILED`.

Here you can see the screenshots I took:

1. First HTTP GET request (just  after creating the report/analysis)
Screenshot from 2019-08-26 17-52-29
Screenshot from 2019-08-26 17-52-53

2. Second HTTP GET request

Screenshot from 2019-08-26 17-53-10
Screenshot from 2019-08-26 17-53-21

NOTE: I've tested it using my local UI environment. It might be difficult to see the problem in CI because it refreshes the page every 5 seconds, I'd suggest you checking the console of your browser to the see problem

File used on test: https://raw.githubusercontent.com/project-xavier/xavier-integration/master/src/test/resources/cloudforms-export-v1.json

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
jonathanvila: the problem is that we could have an exception when we still dont have analysisid in the route ( uploading the file to xavier-integration, or receiving the message from kafka, etc..... )

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
jonathanvila: That's why I created this method.....we can have a list of Success codes that we could configure in the yaml for instance.
But, I will add these other you mention : 201, 202
are you sure about 204 ?

src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
jonathanvila: Honestly, I dont see which is the benefit. We would avoid using the toString but we should implement a class and a method.....
Do you see any harm on having it like this ?

jonathanvila:
> I think we can change the field `status` in the `AnalysisModel` class in order to use ENUMs instead of Strings. @jonathanvila What do you think?

Yes, we can do that.....but I didnt want to touch it as we should investigate the implications on changing the type and what happens in drools, etc.....

jonathanvila:
> @jonathanvila I've just found a problem and I'd like you to verify it, please.
> When a valid `.json` is uploaded, the first status of the analysis is `FAILED` and after a couple of seconds it becomes `CREATED`. I think the first status should be `IN_PROGRESS` and the `CREATED` or `FAILED`.

Seems weird as to have FAILED as the status something should fail in the process. I will take a look.

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
mrizzi: i see your point and when we haven't got the analysis `id` it would fail, thanks

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
mrizzi: I added `204` because we use it so i hope other will use it as well :wink:

mrizzi:
> > I think we can change the field `status` in the `AnalysisModel` class in order to use ENUMs instead of Strings. @jonathanvila What do you think?
> 
> Yes, we can do that.....but I didnt want to touch it as we should investigate the implications on changing the type and what happens in drools, etc.....

I would say i agree with you both :+1: i like the idea of having the `status` to be refactored to not be a `String` but to use the `enum` and, at the same time, i like the idea of the "minimum impact" approach to avoid any unexpected side effect on any other component/application.
I fill a JIRA to keep trace of this proposed change so that we can schedule this later ([MIGENG-214](https://jira.coreos.com/browse/MIGENG-214))

carlosthe19916:
@jonathanvila I've just discovered that the rest endpoint accepts EMPTY String values or WHITE space values for `reportName` field. Is it possible to validate it using `trim()` or something like it? I don't think that empty values on `reportName` should be allowed.

Screenshot from 2019-08-26 18-39-53

jonathanvila:
> @jonathanvila I've just discovered that the rest endpoint accepts EMPTY String values or WHITE space values for `reportName` field. Is it possible to validate it using `trim()` or something like it? I don't think that empty values on `reportName` should be allowed.
> 
> Screenshot from 2019-08-26 18-39-53

The backend can validate that, ..... but, why in the first place the UI allows that ? I would say that the UI should prevent of sending values not "valid".
Also there is no requirement yet on checking that and reject records with report name empty.
@mrizzi do we need to create an issue for this ?

mrizzi:
> > @jonathanvila I've just discovered that the rest endpoint accepts EMPTY String values or WHITE space values for `reportName` field. Is it possible to validate it using `trim()` or something like it? I don't think that empty values on `reportName` should be allowed.
> > Screenshot from 2019-08-26 18-39-53
> 
> The backend can validate that, ..... but, why in the first place the UI allows that ? I would say that the UI should prevent of sending values not "valid".
> Also there is no requirement yet on checking that and reject records with report name empty.
> @mrizzi do we need to create an issue for this ?

I think that having the backend to validate input parameters is something worth having to have safer APIs.
At the same time it was not meant to be done in this PR so good to have discovered it during the review and let's track it in a new issue in JIRA ([MIGENG-215](https://jira.coreos.com/browse/MIGENG-215)) to not forget about it.

carlosthe19916:
> @jonathanvila I've just found a problem and I'd like you to verify it, please.
> When a valid `.json` is uploaded, the first status of the analysis is `FAILED` and after a couple of seconds it becomes `CREATED`. I think the first status should be `IN_PROGRESS` and then `CREATED` or `FAILED`.
> 
> Here you can see the screenshots I took:
> 
> 1. First HTTP GET request (just  after creating the report/analysis)
>    Screenshot from 2019-08-26 17-52-29
>    Screenshot from 2019-08-26 17-52-53
> 2. Second HTTP GET request
> 
> Screenshot from 2019-08-26 17-53-10
> Screenshot from 2019-08-26 17-53-21
> 
> NOTE: I've tested it using my local UI environment. It might be difficult to see the problem in CI because it refreshes the page every 5 seconds, I'd suggest you checking the console of your browser to the see problem
> 
> File used on test: https://raw.githubusercontent.com/project-xavier/xavier-integration/master/src/test/resources/cloudforms-export-v1.json

I've tested the last changes and this problem was solved. Thanks @jonathanvila

carlosthe19916 (approved):
- The problem of incorrect STATUS was solved in the last commit
- The ENUM approach for the status field will be analyzed in the future in another PR [MIGENG-214](https://jira.coreos.com/browse/MIGENG-214)
- The validation of endpoints to not accept empty string values will be done in other PR [MIGENG-215](https://jira.coreos.com/browse/MIGENG-215)

This PR looks good to me and I think we can merge it

src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
mrizzi: https://jira.coreos.com/browse/MIGENG-214

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -58,6 +62,8 @@
     @Inject
     private AnalysisService analysisService;
 
+    private List<Integer> httpSuccessCodes = Arrays.asList(HttpStatus.SC_OK, HttpStatus.SC_ACCEPTED, HttpStatus.SC_NO_CONTENT);
```
mrizzi: Please add also `201` (`SC_CREATED`)
```suggestion
    private List<Integer> httpSuccessCodes = Arrays.asList(HttpStatus.SC_OK, HttpStatus.SC_CREATED, HttpStatus.SC_ACCEPTED, HttpStatus.SC_NO_CONTENT);
```

mrizzi (commented):
Just one further comment to have also `201` (`SC_CREATED`) HTTP code as we agreed.

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -91,19 +97,30 @@ public void configure() {
 
         from("direct:store").id("direct-store")
                 .convertBodyTo(byte[].class) // we need this to fully read the stream and close it
-                .to("file:./upload")
```
mrizzi, line 94: This solves the issue in https://jira.coreos.com/browse/MIGENG-213

mrizzi (approved):
Thanks for this useful enhancement @jonathanvila :+1: 

This PR fixes https://jira.coreos.com/browse/MIGENG-156 and https://jira.coreos.com/browse/MIGENG-213

PR #37, opened 2019-08-19
MIGENG-187: Business Rules Implementation - WSR Workloads Detected

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-187
Added:
- `WorkloadModel`
- `WorkloadService`
- `WorkloadRepository`

Enhanced:
- `direct:calculate-workloadsummaryreportmodel` router
- `WorkloadSummaryReportRoutes_DirectCalculateVMWorkloadInventoryModelTest` test class

PR #46, opened 2019-08-28
MIGENG-231: Business Rules Implementation - WSR Workloads Detected (OS types)

carlosthe19916:
This PR depends on https://github.com/project-xavier/xavier-integration/pull/37

https://jira.coreos.com/browse/MIGENG-231


carlosthe19916:
Latest changes include the creation of WorkloadsDetectedOSTypeModel, its Repository and Service.
The most important change on the last commit is what is written from this line and below:

https://github.com/carlosthe19916/xavier-integration/blob/0719bea9f6948846fa37b46fee3d50578b3ab1ff/src/main/java/org/jboss/xavier/integrations/route/WorkloadSummaryReportRoutes.java#L117

Because the `WorkloadsDetectedOSTypeModel` depends on data from `WorkloadModel` we need to SAVE it in the DB and then execute queries to populate `WorkloadsDetectedOSTypeModel`

carlosthe19916:
Summary of changes:
1. Created `WorkloadsDetectedOSTypeModel` which contains the SQL query to populate the model.
2. Changes made in `WorkloadSummaryReportModel`:
- Changed the collection type of `WorkloadSummaryReportModel#summaryModels`
before: `List<SummaryModel> summaryModels;`
after: `Set<SummaryModel> summaryModels`
**REASON1:** fetching 2 or more `java.util.List` using EAGER will thrown an exception: **org.hibernate.loader.MultipleBagFetchException: cannot simultaneously fetch multiple bags**
**REASON2:** fetching a `WorkloadSummaryReportModel` which have `summaryModels` and `workloadsDetectedOSTypeModels` elements will cause to allways have double elements in `summaryModels` collection. It seems to me that when using `List` and `Set` Hibernate generates different types of queries.

- Changed fetch type to EAGER in @OneToOne relations. => to make it explicit that we want to load them using EAGET

3. Changes made in `AnalysisService` and `WorkloadSummaryReportRoutes`:
`WorkloadsDetectedOSTypeModel` depends on `WorkloadInventoryReportModel` and we need to persist `WorkloadInventoryReportModel` before executing any query to extract `WorkloadsDetectedOSTypeModel`

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/WorkloadsDetectedOSTypeModel.java
```diff
@@ -0,0 +1,102 @@
...
+                columns = {
+                        @ColumnResult(name = "osName", type = String.class),
+                        @ColumnResult(name = "total", type = Integer.class)
+                }
+        )
+)
+
+@NamedNativeQuery(
+        name = "WorkloadsDetectedOSTypeModel.calculateWorkloadsDetectedOSTypeModels",
+        query = "select coalesce(wm.os_name, 'unknown') as osName, count(*) as total \n" +
```
mrizzi: In the `workload_model.vms` column there's the number of times the workload has been detected and it should drive the donut chart as well.
```suggestion
        query = "select coalesce(wm.os_name, 'unknown') as osName, sum(wm.vms) as total \n" +
```

mrizzi (commented):
Change to the query to retrieve the data for the donut chart.

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/WorkloadsDetectedOSTypeModel.java
carlosthe19916: Thanks @mrizzi. I've just added your suggested changes.

mrizzi (approved):
Thanks @carlosthe19916, now the data persisted in the DB looks good :+1:

PR #48, opened 2019-09-02
MIGENG-234: Workload Migration Summary Report - Provider Summary Table

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-234
Alter the SQL query of `SummaryModel`.
- Added `product`, `version`, and `hosts` to the SQL query


mrizzi (commented):
@carlosthe19916 could you update the tests to let them have the new fields as well please?

carlosthe19916:
The last changes include changes from @mrizzi 's branch. Thanks Marco!

Marco's improvements:
- Changed `Sockets` field formula (removed `* 2` because it seemed to be wrong with the latest payload)
- removed `host_name` from `group by` otherwise we would have multiple lines in the Summary table and the `count(distinct host_name)` would have been always `1`
- added `NUMCORESPERSOCKETPATH` field with JSONPath
- changed the `hypervisor` path for the ICS to work with `hosts` array
- changed the JSONPath to get the host name for each VM 

The original message and branch can be found here https://github.com/carlosthe19916/xavier-integration/pull/2#issue-313481061

PR #49, opened 2019-09-02
MIGENG-233 - Flags table mapping for flags and assessment and mappings endpoint

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-233

Add Mapping endpoint which will be used by the "Flag" table to get the equivalent assessment from a specific flag.


carlosthe19916:
added `/mappings/flag-assessment/{flag}` endpoint which will be used by the UI to retrieve the assessment/flag mappings

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/FlagAssessmentModel.java
```diff
@@ -0,0 +1,46 @@
...
+@Table(uniqueConstraints = {
+        @UniqueConstraint(name = "FlagAssessmentModel_" +
+                FlagAssessmentModel.FLAG + "_unique",
+                columnNames = FlagAssessmentModel.FLAG)
+})
+public class FlagAssessmentModel implements java.io.Serializable {
+
+    static final String FLAG = "flag";
+
+    @Id
```
mrizzi: With the unique constraint on the `flag` field i would say this `id` field is useless from the DB point of view and we don't even need it to read the values since `findByFlag` doesn't need it.

mrizzi (changes requested):
During the test, a remote call failed.

src/main/resources/spring/camel-context.xml
```diff
@@ -38,6 +38,15 @@
             <apiProperty key="cors" value="true"/>
         </restConfiguration>
 
+        <rest path="/mappings">
+            <description>Mapping REST API</description>
+            <get uri="/flag-assessment/{flag}">
```
mrizzi: With `>4 vNICs` flag, the `/api/xavier/mappings/flag-assessment/%3E4%20vNICs` endpoint failed returning `400 Bad Request`

mrizzi:
Please refer to also https://jira.coreos.com/browse/MIGENG-233 for detailed description of the requirements

carlosthe19916:
@mrizzi Thanks for reviewing this PR. The first version did not consider the OS as a parameter to find the correct Assessment.

I've considered to maintain the endpoint `api/xavier/mappings/flag-assessment/{flag}` which will return the model below:
> {
>   "flag": "RDM",
>   "label": "Raw Device Mapping",
>   "flagAssessmentOSModels": [
>     {
>       "assessment": "Assessment for null OS"
>     },
>     {
>       "assessment": "Assessment for Linux",
>       "osName": "Linux"
>     },
>     {
>       "assessment": "Assessment for windows",
>       "osName": "Windows"
>     }
>   ]
> }

It means that for every flag, the endpoint will return all possible values for every OS. The UI will be in charge of using the right selection.

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/FlagAssessmentModel.java
carlosthe19916: `flag` is now the PK

src/main/resources/spring/camel-context.xml
carlosthe19916: this problem should not happen again

carlosthe19916:
Last Changes:

- There is a new endpoint available: `api/xavier/mappings/flag-assessment` => will return all the mappings from the DB

The result of the endpoint should be similar to:

> [ { "flag": "RDM", "label": "Raw Device Mapping", "flagAssessmentOSModels": [ { "osName": "", "assessment": "Manual data migration required for Generic OS" }, { "osName": "Windows", "assessment": "Manual data migration required for Windows" }, { "osName": "Linux", "assessment": "Manual data migration required for Linux" } ] } ]

Screenshot from 2019-09-05 16-52-53

- SQL examples to populate the DB

`insert into flag_assessment_model(flag, label) values('RDM', 'Raw Device Mapping');`
`insert into flag_assessmentosmodel(os_name, assessment, flag_assessment_id) values ('Linux', 'Manual data migration required for Linux', 'RDM');`
`insert into flag_assessmentosmodel(os_name, assessment, flag_assessment_id) values ('Windows', 'Manual data migration required for Windows', 'RDM');`
`insert into flag_assessmentosmodel(os_name, assessment, flag_assessment_id) values ('', 'Manual data migration required for Unknown', 'RDM');`

mrizzi (commented):
@carlosthe19916 i gave a look to your last changes.
The models seem to be different from the requirement in https://jira.coreos.com/browse/MIGENG-233.
Now we have a `FlagAssessmentModel` with a 1:N relation with `FlagAssessmentOSModel`.
I don't understand the need for this change in the model and i think it should be better to implement a single table with 4 columns as in the requirement to keep the mapping easy.

carlosthe19916:
> @carlosthe19916 i gave a look to your last changes.
> The models seem to be different from the requirement in https://jira.coreos.com/browse/MIGENG-233.
> Now we have a `FlagAssessmentModel` with a 1:N relation with `FlagAssessmentOSModel`.
> I don't understand the need for this change in the model and i think it should be better to implement a single table with 4 columns as in the requirement to keep the mapping easy.

Thanks @mrizzi for reviewing this PR. As agreed, I've changed the model and now we will just have `FlagAssessmentModel`.

1. This is an example of how we can populate the database:
- `insert into flag_assessment_model(flag,os_name,flag_label,assessment) values ('RDM','Linux', 'Raw Mapping Device', 'Manual data migration required');`

2. Because `flag` and `os_name` are the primary keys, in the database, we can not insert null values there. If we want a generic `os_name` we should insert empty string e.g.
- `insert into flag_assessment_model(flag,os_name,flag_label,assessment) values ('RDM','', 'Raw Mapping Device', 'Generic assessment when no specific OS matches');`

mrizzi (approved):
Thanks  @carlosthe19916: the code now looks good and it has been successfully tested in CI.

PR #50, opened 2019-09-03
Fix: Workload Migration Inventory - search by FlagsIMS

carlosthe19916:
There is a java.lang.NullPointerException when searching by FlagsIMS. It is caused by an error in the source code which is fixed by this PR


mrizzi (approved):
Thanks for this fix @carlosthe19916

PR #52, opened 2019-09-04
Update 'lastUpdate' field when status changes

mrizzi:


mrizzi (commented):
This change needs to be merged to be released in QE so that we can have information about the timing of the analysis

PR #54, opened 2019-09-06
MIGENG-244 Performance improvement

mrizzi:
https://jira.coreos.com/browse/MIGENG-244

- changed the JSON management: instead of parsing it (i.e. `JsonPath.parse(cloudFormsJson);`) many times, we just parse it at the beginning 
- changed the WIR Drools' integration to be `direct` so that, leverage `WorkloadInventoryReportModelAggregationStrategy`, the application can do just one insert for all the WIR entries instead of the previous "one-by-one" approach
- introduced `parallelProcessing` for the generation of the WIR entries with a property to control it (`parallel.wir` with default `true`)


mrizzi (commented):
The issue has been tested in CI env to check the performance was better.

PR #55, opened 2019-09-09
MIGENG-249 - Shared Disks Flag

mrizzi:
https://jira.coreos.com/browse/MIGENG-249
- created `FlagSharedDisksCalculator` and added it after `calculate-vmworkloadinventory` and before WSR creation
- done some refactoring into `AbstractVMWorkloadInventoryCalculator` (`VMWorkloadInventoryCalculator` should be enchanced in another PR)


mrizzi (commented):
Again, i merge my own PR after we tested it in CI env for some days and since we agreed to proceed to merge during a meeting.

PR #57, opened 2019-09-12
MIGENG-257 Access control based on username

mrizzi:
https://jira.coreos.com/browse/MIGENG-257


mrizzi (commented):
This is a blocker issue.
Tested in CI with @PhilipCattanach  and everything worked.
Tests are fine as well.

PR #59, opened 2019-09-27
MIGENG-263 MA - Create CSV exports for analysis metrics

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-263

Changes:
1. Added a new endpoint `/administration/report/csv`, this returns a .zip which includes 2 .csv files as described in the Jira Ticket.
2. Added a new rest interceptor to verify that just allowed users can have access to ´/administration/*´ endpoints. The whitelist of users will come from SpringBoot `application.properties` using the propertie: `rest.authorization.administration`. E.g. `rest.authorization.administration=admin1,admin2@redhat.com`


src/main/resources/application.properties
```diff
@@ -140,3 +140,5 @@ cloudforms.manifest.0.vmworkloadinventory.filesContentPathContents=contents
 cloudforms.manifest.0.vmworkloadinventory.productPath=$.ManageIQ::Providers::Vmware::InfraManager.emstype_description
 cloudforms.manifest.0.vmworkloadinventory.versionPath=$.ManageIQ::Providers::Vmware::InfraManager.api_version
 cloudforms.manifest.0.vmworkloadinventory.hostNamePath=$.ManageIQ::Providers::Vmware::InfraManager.hosts[?(@.id == {host_id})].hostname
+
+rest.authorization.administration=${admin.users}
```
mrizzi: This line prevents this PR to start in CI if there's no `admin.users` env property defined and we have to change this approach.
We need to provide a default so that the application always starts with high security level so change this to become:
```suggestion
rest.authorization.administration=
```
and then we'll take care of providig the list of authorized users in OCP.

src/main/resources/logback.xml
```diff
@@ -12,8 +12,8 @@
 
     <logger name="org.apache.commons.httpclient" level="warn" />
 
-    <root level="info">
+    <root level="warn">
```
mrizzi: It's fine to revert this change
```suggestion
    <root level="info">
```
and manage the Travis CI issue creating a new logback conf file in `/src/test/resources/logback-test.xml` with `root level="warn"`. 
In this way, when running tests, the `logback-test.xml` should prevent any issue in Travis Ci and when running the application the `logback-test.xml` should not be there (it's in the `test` folder) and everything should log as expected.

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
             </get>
         </rest>
 
+        <rest path="/administration" produces="application/octet-stream">
```
mrizzi: I feel like not all of the `/administration` endpoints will produce `application/octet-stream`. Better apply the `produces` attribute to the endpoint it refers to.
```suggestion
        <rest path="/administration">
```

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
...
+            <get uri="/report/csv">
```
mrizzi: See above
```suggestion
            <get uri="/report/csv" produces="application/octet-stream">
```

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
...
+                    </setHeader>
+                    <setHeader headerName="toDate">
+                        <method ref="dateUtils" method="getLastMonday"/>
+                    </setHeader>
+
+                    <log message="Creating csv first file from date ${date:header.fromDate} to ${date:header.toDate}" />
+                    <bean ref="analysisService" method="getAdministrationMetrics(${header.fromDate}, ${header.toDate})"/>
+                    <bean ref="administrationMetricsMapper" method="toAdministrationMetricsModels"/>
+                    <to uri="direct:administrationMetricsModelToCsv"/>
+                    <setHeader headerName="CamelFileName">
```
mrizzi: Use `Exchange.FILE_NAME` instead of its value `CamelFileName`

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
...
+                        <simple>${date:header.fromDate:yyyyMMddHHmmssz}-to-${date:header.toDate:yyyyMMddHHmmssz}.csv</simple>
+                    </setHeader>
+                    <to uri="direct:administration-report-csv-aggregator" />
+
+                    <log message="Creating csv second file from date ${date:header.fromDate} to ${date:header.toDate}" />
+                    <bean ref="analysisService" method="getAdministrationMetrics(${header.unixEpochDate}, ${header.nowDate})"/>
+                    <bean ref="administrationMetricsMapper" method="toAdministrationMetricsModels"/>
+                    <to uri="direct:administrationMetricsModelToCsv"/>
+                    <setHeader headerName="CamelFileName">
```
mrizzi: Use `Exchange.FILE_NAME` instead of its value `CamelFileName`

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
...
+                        <simple>all.csv</simple>
+                    </setHeader>
+                    <to uri="direct:administration-report-csv-aggregator" />
+
+                    <pollEnrich>
+                        <simple>file:tmpFolder?fileName=${header.CSVCorrelationID}.zip</simple>
```
mrizzi: Adding the `delete` parameter would help us in avoiding any useless space usage and to remove the files as soon as they will be consumed.
```suggestion
                        <simple>file:tmpFolder?fileName=${header.CSVCorrelationID}.zip&delete=true</simple>
```

src/main/resources/spring/camel-context.xml
```diff
@@ -208,6 +269,26 @@
...
+            </marshal>
+        </route>
+
+        <route id="administration-report-csv-aggregator">
+            <from uri="direct:administration-report-csv-aggregator"/>
+            <aggregate strategyRef="zipAggregationStrategy" completionSize="2">
+                <correlationExpression>
+                    <header>CSVCorrelationID</header>
+                </correlationExpression>
+                <setHeader headerName="CamelFileName">
```
mrizzi: Use `Exchange.FILE_NAME` instead of its value `CamelFileName`

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
             </get>
         </rest>
 
+        <rest path="/administration" produces="application/octet-stream">
+            <get uri="/report/csv">
+                <route id="administration-report-csv">
+                    <setHeader headerName="CSVCorrelationID">
```
mrizzi: Please create static field in `MainRouteBuilder` (there are already some examples) in order to avoid to repeat the value `CSVCorrelationID` multiple times

src/main/resources/spring/camel-context.xml
mrizzi: See above in order to avoid to repeat the value `CSVCorrelationID` multiple times

src/main/resources/spring/camel-context.xml
```diff
@@ -208,6 +269,26 @@
...
+            <marshal>
+                <bindy id="administrationMetricsModelBindyDataformat" type="Csv" classType="org.jboss.xavier.analytics.pojo.AdministrationMetricsModel"/>
+            </marshal>
+        </route>
+
+        <route id="administration-report-csv-aggregator">
+            <from uri="direct:administration-report-csv-aggregator"/>
+            <aggregate strategyRef="zipAggregationStrategy" completionSize="2">
+                <correlationExpression>
+                    <header>CSVCorrelationID</header>
```
mrizzi: See above in order to avoid to repeat the value `CSVCorrelationID` multiple times

src/main/resources/spring/camel-context.xml
```diff
@@ -208,6 +269,26 @@
...
+                </correlationExpression>
+                <setHeader headerName="CamelFileName">
+                    <simple>${header.CSVCorrelationID}.zip</simple>
```
mrizzi: See above in order to avoid to repeat the value `CSVCorrelationID` multiple times

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +65,53 @@
...
+                    <to uri="direct:administrationMetricsModelToCsv"/>
+                    <setHeader headerName="CamelFileName">
+                        <simple>all.csv</simple>
+                    </setHeader>
+                    <to uri="direct:administration-report-csv-aggregator" />
+
+                    <pollEnrich>
+                        <simple>file:tmpFolder?fileName=${header.CSVCorrelationID}.zip</simple>
+                    </pollEnrich>
+                    <setHeader headerName="Content-Disposition">
```
mrizzi: Better use the `CONTENT_DISPOSITION` field from https://github.com/project-xavier/xavier-integration/blob/master/src/main/java/org/jboss/xavier/integrations/route/dataformat/CustomizedMultipartDataFormat.java#L35

src/main/resources/spring/camel-context.xml
```diff
@@ -29,11 +29,25 @@
...
+
+    <bean id="dateUtils" class="org.jboss.xavier.utils.DateUtils"/>
+    <bean id="utils" class="org.jboss.xavier.utils.Utils"/>
+
     <camelContext id="camel" xmlns="http://camel.apache.org/schema/spring">
 
         <interceptFrom id="interceptor-username" uri="rest:*">
             <to uri="direct:check-authenticated-request" />
         </interceptFrom>
+        <interceptFrom id="interceptor-authorized-request" uri="rest:*">
```
mrizzi: Could this interceptor be defined only for `uri="rest:/administration/*"`  in order to remove the `when` condition?
If not, i see no reason to have this filter and the `when` block could be moved into the previous `interceptor-username` interceptor safely, right?

mrizzi (commented):
I ran a first check on the code and i reported some issues.
Other than my comments, i would like to add:

- there should tests for `isUserAllowedToAdministratorResources` method in `UserServiceTest` to cover the case when `rest.authorization.administration` has values as well as when it has no values. Something like 
```java
    @Test
    public void userService_AuthorizedUsersEmpty_ShouldReturnNotAllowed() {
        ReflectionTestUtils.setField(userService, "authorizedAdminUsers", new String[0]);
        assertThat(userService.isUserAllowedToAdministratorResources("user name")).isEqualTo(false);
    }
```
- it seems like the new route has the block
```xml
<log message="Creating csv first file from date ${date:header.fromDate} to ${date:header.toDate}" />
<bean ref="analysisService" method="getAdministrationMetrics(${header.fromDate}, ${header.toDate})"/>
<bean ref="administrationMetricsMapper" method="toAdministrationMetricsModels"/>
<to uri="direct:administrationMetricsModelToCsv"/>
<setHeader headerName="CamelFileName">
    <simple>${date:header.fromDate:yyyyMMddHHmmssz}-to-${date:header.toDate:yyyyMMddHHmmssz}.csv</simple>
</setHeader>
<to uri="direct:administration-report-csv-aggregator" />
```
that repeats twice: i wonder if there something we can do to avoid duplication and run twice the same block with different dates

src/main/resources/spring/camel-context.xml
carlosthe19916: Thanks, I've fixed this problem.

src/main/resources/spring/camel-context.xml
carlosthe19916: Done

src/main/resources/spring/camel-context.xml
carlosthe19916: Done

src/main/resources/logback.xml
carlosthe19916: Thanks Marco. This is done.

src/main/resources/spring/camel-context.xml
carlosthe19916: done

src/main/resources/spring/camel-context.xml
carlosthe19916: done

src/main/resources/spring/camel-context.xml
carlosthe19916: done

src/main/resources/spring/camel-context.xml
carlosthe19916: `uri="rest:/administration/*"` does not work so I decided to move the code to "interceptor-username"

carlosthe19916:
> I ran a first check on the code and i reported some issues.
> Other than my comments, i would like to add:
> 
> * there should tests for `isUserAllowedToAdministratorResources` method in `UserServiceTest` to cover the case when `rest.authorization.administration` has values as well as when it has no values. Something like
> 
> ```java
>     @Test
>     public void userService_AuthorizedUsersEmpty_ShouldReturnNotAllowed() {
>         ReflectionTestUtils.setField(userService, "authorizedAdminUsers", new String[0]);
>         assertThat(userService.isUserAllowedToAdministratorResources("user name")).isEqualTo(false);
>     }
> ```
> 
> * it seems like the new route has the block
> 
> ```
> <log message="Creating csv first file from date ${date:header.fromDate} to ${date:header.toDate}" />
> <bean ref="analysisService" method="getAdministrationMetrics(${header.fromDate}, ${header.toDate})"/>
> <bean ref="administrationMetricsMapper" method="toAdministrationMetricsModels"/>
> <to uri="direct:administrationMetricsModelToCsv"/>
> <setHeader headerName="CamelFileName">
>     <simple>${date:header.fromDate:yyyyMMddHHmmssz}-to-${date:header.toDate:yyyyMMddHHmmssz}.csv</simple>
> </setHeader>
> <to uri="direct:administration-report-csv-aggregator" />
> ```
> 
> that repeats twice: i wonder if there something we can do to avoid duplication and run twice the same block with different dates

- Test added on UserService
- Refactored `camel-context.xml` in order to avoid duplicate code when creating the csv files

src/main/resources/application.properties
carlosthe19916: Thanks @mrizzi , it is done

src/main/resources/spring/camel-context.xml
carlosthe19916: @mrizzi Regarding to `CSVCorrelationID`. I don't see any value for creating a constant, in a Java Class; this value will be used just on a couple of routes and it has no meaning for the rest of the routes. If we create a constant for `CSVCorrelationID` why don't we create another constant for the rest of custom headers like `fromDate`, `toDate`?, etc. By the way, I've changed `CSVCorrelationID` to `correlationId` just to be consistent with the lowercase pattern of the rest custom headers.

carlosthe19916:
@mrizzi my last commit added dynamic Log Root Level value. By default we will have `info` log level; However, we will use `warn` log level for Travis. In order to change the default log level value we should add `-DlogRootLevel=warn` to maven command.

I think we should keep `info` level on our local environment and use `warn` just for special cases like TravisCI

src/main/resources/spring/camel-context.xml
mrizzi: @carlosthe19916 i see your point and i would say that i agree on having also the other customer headers to use constant values.
The benefit seems low when you use them in 2-3 places (even if it means you changed it from `CSVCorrelationID` to `correlationId` 3 times instead of just 1) but it will help the maintenance of the code since changing once is less error prone than changing it n-times.

mrizzi:
> @mrizzi my last commit added dynamic Log Root Level value. By default we will have `info` log level; However, we will use `warn` log level for Travis. In order to change the default log level value we should add `-DlogRootLevel=warn` to maven command.
> 
> I think we should keep `info` level on our local environment and use `warn` just for special cases like TravisCI

I think it's a good approach: it gives us the freedom to have the level we want in different environments :+1:

pom.xml
```diff
@@ -22,6 +22,9 @@
         <!-- maven plugin versions -->
         <maven-compiler-plugin.version>3.6.0</maven-compiler-plugin.version>
         <maven-surefire-plugin.version>2.19.1</maven-surefire-plugin.version>
+
+        <!--this is used by logback-test.xml-->
+        <logRootLevel>info</logRootLevel>
```
mrizzi: If it affects just the test root logger, better change a little bit the name
```suggestion
        <testLogRootLevel>info</testLogRootLevel>
```

.travis.yml
```diff
@@ -15,6 +15,6 @@ jobs:
   include:
     - stage: Test
       script:
-        - mvn test -Pcoverage
+        - mvn test -DlogRootLevel=warn -Pcoverage
```
mrizzi: See below
```suggestion
        - mvn test -DtestLogRootLevel=warn -Pcoverage
```

src/main/java/org/jboss/xavier/utils/Utils.java
```diff
@@ -0,0 +1,16 @@
+package org.jboss.xavier.utils;
+
+import java.util.Date;
+import java.util.UUID;
+
+public class Utils {
+
+    public Date getUnixEpochDate() {
+        return new Date(0L);
+    }
```
mrizzi: It seems like a duplicate of the same method in `DateUtils`

src/test/java/org/jboss/xavier/integrations/jpa/service/UserServiceTest.java
```diff
@@ -38,4 +38,10 @@ public void userService_AnalysisGiven_ShouldReturnUser() {
         assertThat(userService.findUser("mrizzi@redhat.com").isFirstTimeCreatingReports()).isEqualTo(false);
         assertThat(userService.findUser("whatever").isFirstTimeCreatingReports()).isEqualTo(true);
     }
+
+    @Test
+    public void userService_AuthorizedUsersEmpty_ShouldReturnNotAllowed() {
+        ReflectionTestUtils.setField(userService, "authorizedAdminUsers", new String[0]);
+        assertThat(userService.isUserAllowedToAdministratorResources("myUsername")).isEqualTo(false);
+    }
```
mrizzi, line 46: Better add a test to cover the case when `rest.authorization.administration` has values
```suggestion
    }
    
    @Test
    public void userService_AuthorizedUsersGiven_ShouldReturnAllowedOrNot() {
        assertThat(userService.isUserAllowedToAdministratorResources("admin1")).isEqualTo(true);
        assertThat(userService.isUserAllowedToAdministratorResources("admin1@redhat.com")).isEqualTo(false);
        assertThat(userService.isUserAllowedToAdministratorResources("admin2@redhat.com")).isEqualTo(true);
    }
```

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_InterceptorTest.java
```diff
@@ -0,0 +1,191 @@
...
+    @Value("${camel.component.servlet.mapping.context-path}")
+    String camel_context;
+
+    @Before
+    public void setup() {
+        camel_context = camel_context.substring(0, camel_context.indexOf("*"));
+    }
+
+    @Test
+    public void xmlRouteBuilder_AuthorizedInterceptor_GivenNoRHIdentity_ShouldReturnForbidden() throws Exception {
```
mrizzi: It seems very similar to [`XmlRoutes_RestReportTest.xmlRouteBuilder_RestEndpoints_NoRHIdentityGiven_ShouldReturnForbidden`](https://github.com/project-xavier/xavier-integration/blob/master/src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java#L812) test: i think it worth having the interceptor test in this new `XmlRoutes_InterceptorTest` class so please move the checks from `xmlRouteBuilder_RestEndpoints_NoRHIdentityGiven_ShouldReturnForbidden` into this test/class.

mrizzi (commented):
Added some further comments to the code

src/main/java/org/jboss/xavier/utils/Utils.java
carlosthe19916: you are right, this is definitely a duplicate code. I've fixed it

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_InterceptorTest.java
carlosthe19916: Since both tests were almost similar I've decided to merge them and move it to XmlRoutes_Interceptor.java Class

src/main/resources/spring/camel-context.xml
carlosthe19916: I see your point. My last commit should fix this. Please take a look

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +63,51 @@
...
+                    </setHeader>
+
+                    <setHeader headerName="${type:org.jboss.xavier.integrations.route.MainRouteBuilder.FROM_DATE}">
+                        <method ref="dateUtils" method="getUnixEpochDate"/>
+                    </setHeader>
+                    <setHeader headerName="${type:org.jboss.xavier.integrations.route.MainRouteBuilder.TO_DATE}">
+                        <simple>${date:now}</simple>
+                    </setHeader>
+                    <setHeader headerName="Exchange.FILE_NAME">
+                        <simple>${date:header.${type:org.jboss.xavier.integrations.route.MainRouteBuilder.FROM_DATE}:yyyyMMddHHmmssz}-to-${date:header.${type:org.jboss.xavier.integrations.route.MainRouteBuilder.TO_DATE}:yyyyMMddHHmmssz}.csv</simple>
```
mrizzi: I suppose this should be the `all.csv` file
```suggestion
                        <simple>all.csv</simple>
```

src/main/resources/spring/camel-context.xml
```diff
@@ -51,6 +63,51 @@
...
+                    <to uri="direct:administrationReportCsvGenerator"/>
+
+                    <setHeader headerName="${type:org.jboss.xavier.integrations.route.MainRouteBuilder.FROM_DATE}">
+                        <method ref="dateUtils" method="getLastMondayMinus7days"/>
+                    </setHeader>
+                    <setHeader headerName="${type:org.jboss.xavier.integrations.route.MainRouteBuilder.TO_DATE}">
+                        <method ref="dateUtils" method="getLastMonday"/>
+                    </setHeader>
+                    <setHeader headerName="Exchange.FILE_NAME">
+                        <simple>all.csv</simple>
```
mrizzi: I think this should be the report with the dates in the file name
```suggestion
                        <simple>${date:header.${type:org.jboss.xavier.integrations.route.MainRouteBuilder.FROM_DATE}:yyyyMMddHHmmssz}-to-${date:header.${type:org.jboss.xavier.integrations.route.MainRouteBuilder.TO_DATE}:yyyyMMddHHmmssz}.csv</simple>
```

src/main/java/org/jboss/xavier/analytics/pojo/AdministrationMetricsModel.java
```diff
@@ -0,0 +1,77 @@
...
+    @DataField(pos = 2, columnName = "Domain")
+    private String ownerDomain;
+
+    @DataField(pos = 3, columnName = "Payload name")
+    private String payloadName;
+
+    @DataField(pos = 4, columnName = "Analysis status")
+    private String analysisStatus;
+
+    @DataField(pos = 5, columnName = "Creation date (yyyy-MM-dd)", pattern = "yyyy-MM-dd")
```
mrizzi: could you add the time as well please beside the date?

mrizzi (commented):
I managed to test this and i found just one issue with file names.
I talked with @PhilipCattanach and i would be nice to have the time in the dates

src/main/java/org/jboss/xavier/analytics/pojo/AdministrationMetricsModel.java
carlosthe19916: I've added the time and timeZone using the format `yyyy-MM-dd HH:mm:ss z`. E.g. `2019-09-25 11:51:42 UTC`

mrizzi (approved):
Thanks  @carlosthe19916 for the great effort in getting this PR done :+1:

PR #61, opened 2019-10-01
Migeng 243 exception handling

jonathanvila:
https://jira.coreos.com/browse/MIGENG-243


src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
```diff
@@ -117,4 +119,13 @@ public void updateStatus(String status, Long id) {
         analysisModel.setLastUpdate(new Date());
         analysisRepository.save(analysisModel);
     }
+
+    public void markAsFailedIfNotCreated(Long id) {
+        AnalysisModel analysisModel = findById(id);
+        if (!STATUS.CREATED.toString().equalsIgnoreCase(analysisModel.getStatus())) {
```
mrizzi: Maybe it would better to apply the selection criteria directly in the query with something like `findByIdAndStatusIsNot(Long id, String status)` so that we can leverage the SQL condition.

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
+package org.jboss.xavier.integrations.route;
+
+import org.apache.camel.Exchange;
+import org.apache.camel.builder.RouteBuilder;
+import org.jboss.xavier.integrations.jpa.service.AnalysisService;
+
+import javax.inject.Inject;
+import java.util.Map;
+
+import static org.jboss.xavier.integrations.route.MainRouteBuilder.ANALYSIS_ID;
```
mrizzi: Could we refactor all the static fields from `MainRouteBuilder` to `RouteBuilderExceptionHandler` in order to avoid to have this class to import from one of the class that extends it?

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public void markAnalysisAsFailed(Exchange e) {
+        try {
+            String analysisId = e.getIn().getHeader(ANALYSIS_ID, "", String.class);
+            if (analysisId.isEmpty()) {
+                analysisId = (String) e.getIn().getHeader(MA_METADATA, Map.class).get(ANALYSIS_ID);
+            }
+            analysisService.markAsFailedIfNotCreated(Long.parseLong(analysisId));
+        } catch (Exception ex) {
+            log.error("Exception ocurred while marking the Analysis as failed.", ex);
```
mrizzi: typo
```suggestion
            log.error("Exception occurred while marking the Analysis as failed.", ex);
```

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+                .stop();
+    }
+
+    public void markAnalysisAsFailed(Exchange e) {
+        try {
+            String analysisId = e.getIn().getHeader(ANALYSIS_ID, "", String.class);
+            if (analysisId.isEmpty()) {
+                analysisId = (String) e.getIn().getHeader(MA_METADATA, Map.class).get(ANALYSIS_ID);
+            }
+            analysisService.markAsFailedIfNotCreated(Long.parseLong(analysisId));
```
mrizzi: It seems like there's no need to check for `analysisId` to not be null which is good if we're sure it has always a value. Is this true?

mrizzi (commented):
Run a code review before testing.
Added some comments.

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
jonathanvila: I'll add the check... seems unlikely to have MA_METADATA and not the ANALYSIS_ID, but better to be sure

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
```diff
@@ -0,0 +1,45 @@
...
+    public static final String CORRELATION_ID = "correlationId";
+    public static final String WORKING_DIR = "workingFile";
+    public static final String WORKING_FILE = "workingFile";
+    public static final String FROM_DATE = "fromDate";
+    public static final String TO_DATE = "toDate";
+    @Inject
+    protected AnalysisService analysisService;
+
+    @Override
+    public void configure() throws Exception {
```
mrizzi, line 25: I think it would be good to have a log at level `ERROR` so that we can find in the log the exception and the ID of analysis (if available)

src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
```diff
@@ -123,4 +125,17 @@ public void updateStatus(String status, Long id) {
...
+    public void markAsFailedIfNotCreated(Long id) {
+        AnalysisModel analysisModel = findByIdAndStatusIgnoreCaseNot(id, STATUS.CREATED.toString());
+        if (analysisModel != null) {
+            analysisModel.setStatus(STATUS.FAILED.toString());
+            analysisModel.setLastUpdate(new Date());
+            analysisRepository.save(analysisModel);
+        }
+    }
+
+    private AnalysisModel findByIdAndStatusIgnoreCaseNot(Long id, String status) {
```
mrizzi: Worth adding a unit test for this new method in `AnalysisServiceTest`

mrizzi (commented):
Just a test and some log and then it'll fine for me.

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
jonathanvila: ok

src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
```diff
@@ -0,0 +1,47 @@
...
+        String analysisId = "";
+        try {
+             analysisId = e.getIn().getHeader(ANALYSIS_ID, "", String.class);
+            if (analysisId.isEmpty()) {
+                analysisId = (String) e.getIn().getHeader(MA_METADATA, Map.class).get(ANALYSIS_ID);
+            }
+            if (StringUtils.isNotEmpty(analysisId)) {
+                analysisService.markAsFailedIfNotCreated(Long.parseLong(analysisId));
+            }
+            log.error("Exception occurred while running the Analysis [{}]", analysisId);
```
mrizzi: It would be better to have the exception with its stacktrace in the log so that we can know what caused the failure and be able to do a better analysis of the issue.

mrizzi (commented):
One further comment on exception handling to improve our analysis capabilities.

src/main/java/org/jboss/xavier/integrations/route/RouteBuilderExceptionHandler.java
jonathanvila: ok

mrizzi (approved):
Tested in CI creating different failure situation and every analysis failed as expected (meaning that the analysis's failure was expected, managed and reflected in the status).
Thanks @jonathanvila for this enhancement to analysis workflow :+1:

PR #62, opened 2019-10-01
MIGENG-252: ui wsr flag box default sort order incorrect

m-brophy:


carlosthe19916:
I've tested the last changes and it worked well. I compared the result of a SQL query and the actual result of the endpoint and they matched. The default order will be the insertion order of the table.

carlosthe19916 (approved):

PR #63, opened 2019-10-03
Migeng 236 scan run date field

jonathanvila:
https://jira.coreos.com/browse/MIGENG-236

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
+++ b/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
@@ -10,6 +10,7 @@ import javax.inject.Inject;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -56,6 +57,7 @@ public class AbstractVMWorkloadInventoryCalculator {
 
     protected DocumentContext jsonParsed;
     protected String manifestVersion;
+    protected Date scanRunDate;
 
     protected Map<String, String> readMapValuesFromExpandedEnvVarPath(String envVarPath, Map vmStructMap, String keyfield, String valuefield) {
         String expandParamsInPath = getExpandedPath(envVarPath, vmStructMap);
--- a/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
+++ b/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
@@ -1,6 +1,7 @@
 package org.jboss.xavier.integrations.migrationanalytics.business;
 
 import com.jayway.jsonpath.DocumentContext;
+import lombok.extern.slf4j.Slf4j;
 import org.jboss.xavier.integrations.migrationanalytics.business.versioning.ManifestVersionService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.env.Environment;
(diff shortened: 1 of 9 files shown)


src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
```diff
@@ -68,14 +69,20 @@
...
+            value = jsonParsed.read(expandParamsInPath);
+            if (value instanceof Collection) {
+                value = ((List<T>) value).get(0);
+            }
+            if (Long.class.isAssignableFrom(type)) {
+                value = ((Number) value).longValue();
+            } else if (Integer.class.isAssignableFrom(type)) {
+                value = ((Number) value).intValue();
+            }
+        } catch (Exception e) {
```
mrizzi, line 89: Maybe worth adding a log at warning level so that it's clear doesn't work as expected?

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
```diff
@@ -56,6 +59,14 @@ private VMWorkloadInventoryModel createVMWorkloadInventoryModel(Map vmStructMap,
         model.setVersion(readValueFromExpandedEnvVarPath(VERSIONPATH, vmStructMap));
         model.setHost_name(readValueFromExpandedEnvVarPath(HOSTNAMEPATH, vmStructMap));
 
+        Date scanRunDate;
+        try {
+            scanRunDate = new SimpleDateFormat("yyyy-M-dd'T'hh:mm:ss.S").parse(readValueFromExpandedEnvVarPath(DATACOLLECTEDON, vmStructMap));
+        } catch (Exception e) {
+            scanRunDate = new Date();
```
mrizzi: Also here i think it's worth having a warn log to say we set to `now` the date because we have not been able to read it due to `e`

mrizzi (commented):
Other then the comments, i think it's worth adding 
`expectedModel.setScanRunDate(new SimpleDateFormat("yyyy-M-dd'T'hh:mm:ss.S").parse("2019-09-18T14:52:45.871Z"));`
in https://github.com/project-xavier/xavier-integration/blob/master/src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_DirectCalculateVMWorkloadInventoryTest.java#L133 so that the `expectedModel` has the new field

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
```diff
@@ -56,6 +61,15 @@ private VMWorkloadInventoryModel createVMWorkloadInventoryModel(Map vmStructMap,
         model.setVersion(readValueFromExpandedEnvVarPath(VERSIONPATH, vmStructMap));
         model.setHost_name(readValueFromExpandedEnvVarPath(HOSTNAMEPATH, vmStructMap));
 
+        Date scanRunDate;
+        try {
```
mrizzi: Running some test in CI, i figured out that `data_collected_on` is a field that we could read once in `calculate` method instead of reading it every time for every VM.
This helps the case when the field is available (since we read it just once) and maybe even more in the case when it's not available and we have to manage two exceptions each time we create a VM instead of just managing the 2 exceptions once.

mrizzi (commented):
Added one further comment after some tests in CI to improve the performance for the new `data_collected_on` collection.

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
jonathanvila: ok

mrizzi (approved):
Tested in CI and everything works.
Thanks @jonathanvila for this improvement to payload management :+1:

PR #65, opened 2019-11-07
MIGENG-279: 'Export as CSV' does not export filtered result

carlosthe19916:
https://jira.coreos.com/browse/MIGENG-279

Added a new endpoint `/report/{id}/workload-inventory/filtered-csv` which will return a CSV file. It will use the query parameters sent to the endpoint to filter the results.


src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
```diff
@@ -49,6 +49,23 @@
         return reportRepository.findByAnalysisOwnerAndAnalysisId(analysisOwner, analysisId);
     }
 
+    public List<WorkloadInventoryReportModel> findByAnalysisOwnerAndAnalysisId(
```
jonathanvila, line 56: I can't see any Test testing the body of this method.

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
```diff
@@ -49,6 +49,23 @@
...
+            String analysisOwner,
+            Long analysisId,
+            SortBean sortBean,
+            WorkloadInventoryFilterBean filterBean
+    ) {
+        // Sort
+        Sort.Direction sortDirection = sortBean.isOrderAsc() ? Sort.Direction.ASC : Sort.Direction.DESC;
```
jonathanvila: If we dont specify anything it will fall-back to DESC sorting. Is that what it is expected ? I would expect an ASC sorting by default

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
carlosthe19916: Added a test

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
carlosthe19916: @jonathanvila I've read the documentation from Miguel and what is required is to have Default order by: `Provider, Datacenter, Cluster, VM name`. I've made some changes to be consistent with what is required from the Document.

https://docs.google.com/document/d/1jTPrwKFdVIO7bkhNLA6skIneQYBmReaJQDlJW9SoZZk/edit#

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
```diff
@@ -520,6 +520,156 @@ public void xmlRouteBuilder_RestReportIdWorkloadInventory_IdParamGiven_ShouldCal
...
+        camelContext.startRoute("filtered-workload-inventory-report-get-details-as-csv");
+        camelContext.startRoute("workload-inventory-report-model-to-csv");
+
+        ResponseEntity<String> response = restTemplate.exchange(camel_context + "report/{id}/workload-inventory/filtered-csv", HttpMethod.GET, entity, String.class, variables);
+
+        //Then
+        SortBean sortBean = new SortBean("id", false);
+        WorkloadInventoryFilterBean filterBean = new WorkloadInventoryFilterBean();
+
+        verify(workloadInventoryReportService).findByAnalysisOwnerAndAnalysisId("mrizzi@redhat.com", one, sortBean, filterBean);
```
jonathanvila: We could take advantage that we have the Spring context up and H2 and check that really the values filtered by Hibernate are the ones we expect.

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
```diff
@@ -520,6 +520,156 @@ public void xmlRouteBuilder_RestReportIdWorkloadInventory_IdParamGiven_ShouldCal
...
+        filterBean.setClusters(new HashSet<>(Arrays.asList(cluster1, cluster2)));
+        filterBean.setDatacenters(new HashSet<>(Arrays.asList(datacenter1, datacenter2)));;
+        filterBean.setVmNames(new HashSet<>(Arrays.asList(vmName1, vmName2)));;
+        filterBean.setOsNames(new HashSet<>(Arrays.asList(osName1, osName2)));;
+        filterBean.setWorkloads(new HashSet<>(Arrays.asList(workload1, workload2)));
+        filterBean.setRecommendedTargetsIMS(new HashSet<>(Arrays.asList(recommendedTarget1, recommendedTarget2)));
+        filterBean.setFlagsIMS(new HashSet<>(Arrays.asList(flag1, flag2)));
+        filterBean.setComplexities(new HashSet<>(Arrays.asList(complexity1, complexity2)));
+
+        verify(workloadInventoryReportService).findByAnalysisOwnerAndAnalysisId("mrizzi@redhat.com", one, sortBean, filterBean);
```
jonathanvila: We could take advantage that we have the Spring context up and H2 and check that really the values sorted by Hibernate are the ones we expect.

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
```diff
@@ -520,6 +530,266 @@ public void xmlRouteBuilder_RestReportIdWorkloadInventory_IdParamGiven_ShouldCal
...
+        assertThat(response).isNotNull();
+        assertThat(response.getBody()).contains("Provider,Datacenter,Cluster,VM name,OS type,Operating system description,Disk space,Memory,CPU cores,Workload,Effort,Recommended targets,Flags IMS,Product,Version,HostName");
+
+//        Expected CSV result
+//        ProviderA | DatacenterC | ClusterA | VmNameC
+//        ProviderB | DatacenterA | ClusterB | VmNameA
+//        ProviderB | DatacenterB | ClusterB | VmNameB
+        assertThat(response.getBody()).contains("ProviderA,DatacenterC,ClusterA,VmNameC");
+        assertThat(response.getBody()).contains("ProviderB,DatacenterA,ClusterB,VmNameA");
+        assertThat(response.getBody()).contains("ProviderB,DatacenterB,ClusterB,VmNameB");
```
jonathanvila: If I'm not wrong you are checking that the body containes those 3 texts, but not the order of them , as every assert will return True no matter the order they have.
Is that what you wanted ?

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
```diff
@@ -520,6 +530,266 @@ public void xmlRouteBuilder_RestReportIdWorkloadInventory_IdParamGiven_ShouldCal
...
+        WorkloadInventoryFilterBean filterBean = new WorkloadInventoryFilterBean();
+        filterBean.setDatacenters(new HashSet<>(Arrays.asList(datacenter1, datacenter2)));
+
+        verify(workloadInventoryReportService).findByAnalysisOwnerAndAnalysisId(analysisModel.getOwner(), analysisModel.getId(), sortBean, filterBean);
+        Assert.assertTrue(response.getHeaders().get("Content-Type").contains("text/csv"));
+        Assert.assertTrue(response.getHeaders().get("Content-Disposition").contains("attachment;filename=workloadInventory_1.csv"));
+        assertThat(response).isNotNull();
+        assertThat(response.getBody()).contains("Provider,Datacenter,Cluster,VM name,OS type,Operating system description,Disk space,Memory,CPU cores,Workload,Effort,Recommended targets,Flags IMS,Product,Version,HostName");
+        assertThat(response.getBody()).contains("my providerA,my datacenter2");
+        assertThat(response.getBody()).contains("my providerB,my datacenter1");
```
jonathanvila, line 683: I assume your goal was to check that "provider" is not on the list, but you are only checking the ones thar are in the list but not the not which should not be in the list.
Am I wrong ?

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
carlosthe19916: @jonathanvila thanks for noticing that... I've added a new commit where I'm splitting the response's body into rows and then check the order.

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
carlosthe19916: @jonathanvila Thanks for your review, my last commit added a new assert to make sure we have only 3 rows (including the headers) so we know there are no extra rows

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
carlosthe19916: @jonathanvila as you can see in this line https://github.com/carlosthe19916/xavier-integration/blob/443b2364733bf2ffab182329e08454b07b4f12be/src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java#L611 and in the lines below it, there is an `analysisModel` and 3 `worloadInventoryReportModel` which are created in the DB. The final asserts of the method use the DB data

src/test/java/org/jboss/xavier/integrations/route/XmlRoutes_RestReportTest.java
carlosthe19916: @jonathanvila now the tests are is including insertions in the DB and use it in the asserts

jonathanvila:
Code changes OK
CI Tested  OK

Provider | Datacenter | Cluster | VM name
------------|----------------|-----------|---------------
vSphere | JON TEST DC | VMCluster | db
vSphere | JON TEST DC | VMCluster | hana
vSphere | JON TEST DC | VMCluster | jboss0
vSphere | JON TEST DC | VMCluster | jboss1
vSphere | JON TEST DC | VMCluster | lb
vSphere | JON TEST DC | VMCluster | mssql_db
vSphere | JON TEST DC | VMCluster | oracle_db
vSphere | JON TEST DC | VMCluster | tomcat

PR #66, opened 2019-12-12
Release Stage

carlosthe19916:
We need to think about the release process of Xavier repositories. Here is a draft we can use in order to have our release workflow completed.
What this PR does:
- When a new tag is identified, Travis will create a new Release for us. However, we need to create and add a GITHUB_OAUTH_TOKEN into Travis settings using Travis` webpage.
- When a new tag is identified, Travis will create docker images and push them to Docker Hub. We need to create and add DOCKER_USERNAME and DOCKER_PASSWORD (can be a generated token) to Travis.

Suggestions:
- We can build the docker images and push them to Quay.io as well, it is just a matter of adding additional commands


jonathanvila:
Should we rethink this release process trying to align it with what's defined in document for MA going GA ? @mrizzi

PR #67, opened 2020-01-07
MIGENG-324: MA - Prototype RBAC

carlosthe19916:
https://issues.redhat.com/browse/MIGENG-324

This PR implements a very basic prototype about the RBAC control access

Note: Complex version of RBAC was saved here:
https://github.com/carlosthe19916/xavier-integration/tree/MIGENG-324-save-code


src/main/java/org/jboss/xavier/integrations/rbac/RBACService.java
```diff
@@ -0,0 +1,265 @@
...
+            String resourceType = permComponents[1];
+            String operation = permComponents[2];
+
+            List<String> resources = extractResourceDefinitions(resourceDefinitions);
+            AclData aclData = new AclData(operation, resources);
+
+            if (!access.containsKey(resourceType)) {
+                access.put(resourceType, new ArrayList<>());
+            }
+            access.get(resourceType).add(aclData);
```
mrizzi: We can rely upon Map's `computeIfAbsent` method to manage the "missing key case"
```suggestion
            access.computeIfAbsent(resourceType, resource -> new ArrayList<>()).add(aclData);
```

src/main/java/org/jboss/xavier/integrations/rbac/Acl.java
```diff
@@ -0,0 +1,90 @@
+package org.jboss.xavier.integrations.rbac;
+
+import java.util.List;
+
+public class Acl {
```
mrizzi, line 24: could you add comments to describe this bean and its relationship with the `AclData` bean please?

src/main/java/org/jboss/xavier/integrations/rbac/AclData.java
```diff
@@ -0,0 +1,33 @@
+package org.jboss.xavier.integrations.rbac;
+
+import java.util.List;
+
+public class AclData {
```
mrizzi: could you add comments to describe this bean and its relationship with the `Acl` bean please?

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+                .process(exchange -> {
+                    // save decoded x-rh-identity JsonNode as header
+                    String xRHIdentity = exchange.getIn().getHeader(RBAC_X_RH_IDENTITY, String.class);
+                    JsonNode xRHIdentityDecodedJsonNode;
+                    try {
+                        xRHIdentityDecodedJsonNode = new ObjectMapper().reader().readTree(
+                                new String(
+                                        Base64.getDecoder().decode(xRHIdentity)
+                                )
+                        );
```
mrizzi: It seems like a duplication of the [`getRHIdentity`](https://github.com/project-xavier/xavier-integration/blob/master/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java#L349) method used in https://github.com/project-xavier/xavier-integration/blob/master/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java#L123.
Could you please refactor the code in oorder to have just one implementation used everywhere to decode this header?

mrizzi (commented):
@carlosthe19916 i did a partial code review that i want to share with you in order to have working on this while i keep processing the PR for review.

src/main/java/org/jboss/xavier/integrations/rbac/RBACService.java
carlosthe19916: Thanks @mrizzi

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+
+                            oldExchange.getIn().setHeader(RBAC_USER_ACCESS, accessForUser);
+                            return oldExchange;
+                        })
+                    .endChoice()
+                .end();
+
+        from("direct:fetch-rbac-user-access")
+                .routeId("fetch-rbac-user-access")
+                .setHeader("access", ArrayList::new)
```
mrizzi: Please make `access` a constant

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+                .setHeader("nextLink", () -> "")
```
mrizzi: Please make `nextLink` a constant

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+                .loopDoWhile(exchange -> exchange.getIn().getHeader("nextLink") != null)
+                    .setHeader(Exchange.HTTP_METHOD, constant(HttpMethods.GET))
+                    .setHeader(Exchange.CONTENT_TYPE, constant("application/json"))
+                    .setHeader(Exchange.HTTP_PATH, constant(rbacPath))
+                    .process(exchange -> {
+                        String httpQuery;
+                        String nextLink = exchange.getIn().getHeader("nextLink", String.class);
+                        int queryParamsIndex = nextLink.indexOf("?");
```
mrizzi: For just one character there's the "dedicated" `indexOf(int ch)` method
```suggestion
                        int queryParamsIndex = nextLink.indexOf('?');
```

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+                        if (queryParamsIndex != -1) {
+                            httpQuery = nextLink.substring(queryParamsIndex + 1);
+                        } else {
+                            httpQuery = "application=" + rbacApplicationName + "&limit=100&offset=0";
```
mrizzi, line 77: It seems like `limit` query param is ignored from remote system since the response payload has the `1000` value
```
"first": "/api/rbac/v1/access/?application=migration-analytics&limit=1000&offset=0"
```
Could we remove it?
```suggestion
                            httpQuery = "application=" + rbacApplicationName + "&offset=0";
```
Other than this, since `rbacApplicationName` is coming from a property read at boot time and then never changed, i feel like this initial HTTP query could be created just once instead of in every call.
Something like defining inside `configure()` before `from("direct:fetch-rbac-user-access")`
```
final String initialHttpQuery = "application=" + rbacApplicationName + "&offset=0";
```
and then use it in the route.

mrizzi (commented):
another partial review for adding constants for recurring strings (`access` and `nextLink`) and some suggestions for optimizations based on the fact that these routes are going to be used always and hence performance is very important

src/main/java/org/jboss/xavier/integrations/rbac/Acl.java
carlosthe19916: @mrizzi added some docs, let me know if there is something missing or it needs to be enhanced.

src/main/java/org/jboss/xavier/integrations/rbac/AclData.java
carlosthe19916: @mrizzi added some docs, let me know if there is something missing or it needs to be enhanced.

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
carlosthe19916: @mrizzi it seems that the default `Page Size` for that endpoint has been increased to `1000`, if my memory is working well, before it used to be just `10`. I suggest to keep the `limit` query parameter but increase it to `1000`. Keeping the `limit` query parameter will allow us to be in control of the `Page Size`, otherwise we will rely on default values that might change in the RBAC side (I don't think they will change it often but the possibility is still there). Wdyt?

mrizzi: @carlosthe19916 My concern is exactly this one: we are *not* in control at all since, right now, we have `100` and the RBAC response has `1000` totally discarding our request to just have `100` so i feel like this is not really a query param maybe on their side.
Since it's useless (we request `100` and it gave use `1000`) i think it should be removed since it gives the wrong perception of being able to control it.

carlosthe19916: @mrizzi now I'm curious how you got the `1000` number even though you used `limit=100`, maybe you discovered some bug in my code or in the RBAC side... I've tried the OpenAPI interface and this is what i got:

Screenshot from 2020-03-10 12-49-23

As you can see the rest endpoint is using `limit=88`

mrizzi: Sorry @carlosthe19916, it has been a mistake i made in doing the remote call from terminal to simulate the RBAC calls.
Everything works as expected :+1: 
I think `limit=100` is fine

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
carlosthe19916: done

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
carlosthe19916: done

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
carlosthe19916: done

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
carlosthe19916: @mrizzi refactored in this commit: https://github.com/project-xavier/xavier-integration/pull/67/commits/bf59b75f275b840aa80a3a12cbaf4293bc1756dd

Created a static `Processor` which should be in charge of adding all headers related to `x-rh-identity`

src/main/java/org/jboss/xavier/integrations/rbac/RBACService.java
```diff
@@ -0,0 +1,262 @@
...
+    public static Map<String, List<AclData>> processAcls(List<Acl> acls) {
+        Map<String, List<AclData>> access = new HashMap<>();
+        for (Acl acl : acls) {
+            String permission = acl.getPermission();
+            List<Acl.ResourceDefinition> resourceDefinitions = acl.getResourceDefinitions();
+
+            // extract permission_data
+            String[] permComponents = permission.split(":");
+            if (permComponents.length != 3) {
+                throw new IllegalStateException("Invalid permission definition permission:" + permission);
```
mrizzi:
```suggestion
                throw new IllegalStateException("Invalid permission definition:" + permission);
```

src/main/java/org/jboss/xavier/integrations/rbac/RBACService.java
```diff
@@ -0,0 +1,262 @@
+package org.jboss.xavier.integrations.rbac;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class RBACService {
```
mrizzi: This seems to be more a `RBACUtils` with static methods
```suggestion
public class RBACUtils {
```

mrizzi (commented):
Two minor wording issue

src/main/java/org/jboss/xavier/integrations/rbac/RBACService.java
carlosthe19916: renamed

carlosthe19916:
- Last changes removed all the complex code for processing `resourceDefinitions`. We are going to omit the content of `resourceDefinitions` until we have a clear idea about the scenarios that we want to cover.

- The previous code was saved in https://github.com/carlosthe19916/xavier-integration/tree/MIGENG-324-save-code in case we need it in the future.

src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_DirectAddUsernameHeaderTest.java
```diff
@@ -43,7 +43,7 @@ public void mainRouteBuilder_routeDirectAddUsernameHeader_NoContentGiven_ShouldA
         });
 
         //Then
-        assertThat(result.getIn().getHeader("analysisUsername")).isEqualTo("");
+        assertThat(result.getIn().getHeader("analysisUsername")).isEqualTo(null);
```
mrizzi: Please move it to be (also in the cases below):
```suggestion
        assertThat(result.getIn().getHeader("analysisUsername")).isNull();
```

src/main/resources/spring/camel-context.xml
```diff
@@ -222,6 +223,9 @@
                 <route id="workload-inventory-report-get-details-as-csv">
                     <bean ref="workloadInventoryReportService" method="findByAnalysisOwnerAndAnalysisId(${header.${type:org.jboss.xavier.integrations.route.RouteBuilderExceptionHandler.USERNAME}}, ${header.id})" />
                     <to uri="direct:toWorkloadInventoryReportCsv" />
+                    <setHeader headerName="${type:org.jboss.xavier.integrations.route.dataformat.CustomizedMultipartDataFormat.CONTENT_TYPE}">
+                        <constant>text/csv</constant>
+                    </setHeader>
```
mrizzi: Is this needed in this PR?

mrizzi (commented):
Two minor comments

src/main/resources/spring/camel-context.xml
carlosthe19916: it is not needed... I've removed it

mrizzi (approved):
Thanks @carlosthe19916 for the RBAC integration: it has been a challenging task but i hope also a nice one :+1:

PR #68, opened 2020-01-14
Migeng 327 avoid crashes based on issues use cases

jonathanvila:
There are 3 issues covered inside : 
* VM without HOST
* HOST without CLUSTER
* HOST or VM without Total CPUs

These issues can happen because the attribute doesnt exist , or because it has the wrong value type.
Also this PR includes a Handler used to log the details of those Exceptions , but can be extended to persite the info in DB, etc.
Every use case has a json file in resources folder and EndToEnd test uses them to execute the analysis and then asses the values received.

This PR is blocked by https://issues.redhat.com/browse/MIGENG-328 , as there are some changes to do in xavier-analytics to relax the filtering out of vms if certain fields are null


src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/AbstractVMWorkloadInventoryCalculator.java
```diff
@@ -100,11 +104,16 @@
     protected <T> List<T> readListValuesFromExpandedEnvVarPath(String envVarPath, Map vmStructMap) {
         String expandParamsInPath = getExpandedPath(envVarPath, vmStructMap);
```
m-brophy: A very minor point but...Just for readability and clarity, it might be a good idea to rename this local variable so it's not the same as the method 15-20 lines further down in the class.

jonathanvila: +1

m-brophy (approved):

m-brophy (approved):

PhilipCattanach:
Tested in CI with AllScripts payload and a test file that had manually added data integrity issues and the payload were successfully uploaded with sensible defaults.

PR #69, opened 2020-01-16
MIGENG-371: Model inconsistency: More than one row with the given identifier was found

carlosthe19916:
Fixing the Model inconsistency error.


carlosthe19916:
@mrizzi This PR is no longer applicable since https://github.com/project-xavier/xavier-integration/pull/76 has changed the fetch method from EAGER to LAZY. I think we can close this PR without merging it.

carlosthe19916:
Closing PR since is no longer applicable

PR #71, opened 2020-01-22
hotfix-kafka-autooffset set autoOffsetReset to latest

jonathanvila:
Restored the old value that this property had : autoOffsetReset=latest on the kafka camel component to avoid re consuming mensages from kafka and having duplications on DB


mrizzi (approved):
Tested in CI

1. built & deployed hotfix in CI
2. ran 4 analysis
3. scale to 0 and then to 1
4. no duplications
5. ran 2 more analysis
6. scale to 0 and then to 1
7. no duplications

In "restarts" in step 3 and 6 i took the offsets values and they change accordingly.

PR #74, opened 2020-02-04
Migeng 412 big files break app

jonathanvila:
https://issues.redhat.com/browse/MIGENG-412
Changed from **direct** Camel component to **seda**  :
* queue size by default = 1000 messages
* concurrent consumers by default = 1
* block or discard when queue full = false ( that means if a new message arrives when we have 1000 messages waiting, it will throw an exception and be handled by the ExceptionHandler )

Added a **Thread Pool** with a default value of 5 ( env var configurable **thread.pool-size**)

Tested with 2 users ( Phil and me ) uploading big files and small files and small files are not affected by the big files on the analysis.


src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
```diff
@@ -125,10 +128,10 @@ public void configure() throws Exception {
                 .routeId("kafka-upload-message")
                 .unmarshal().json(JsonLibrary.Jackson, FilePersistedNotification.class)
                 .filter(simple("'{{insights.service}}' == ${body.getService}"))
-                .to("direct:download-file");
+                .to("seda:download-file");
 
-        from("direct:download-file")
-                .routeId("download-file")
+        from("seda:download-file").routeId("download-file")
+                .threads(THREAD_POOL_SIZE)
```
mrizzi: I took some time to understand how this works.
It seems that [SEDA component doc](https://camel.apache.org/components/latest/seda-component.html#_thread_pools) suggests to beware about using thread with seda and to prefer the `concurrentConsumers` option.
So i think we should follow this warning and move to `concurrentConsumers` option even if this means having a fixed size pool.

jonathanvila:
I do not agree

concurrentConsumers is an approach that will use FIXED number of consumers
as it is said here :
"As for the difference between the two, note a *thread pool* can
increase/shrink dynamically at runtime depending on load, whereas the
number of concurrent consumers is always fixed"

src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java
jonathanvila: concurrentConsumers is an approach that will use FIXED number of consumers
as it is said here :
"As for the difference between the two, note a *thread pool* can
increase/shrink dynamically at runtime depending on load, whereas the
number of concurrent consumers is always fixed"

jonathanvila: In my opinion we always going to have 1 blocking queue from the SEDA component, and with the threads we are adding another one dynamic. If we decide to increase from 5 to 20 , with the concurrentConsumers will be always having 20 consumers , but not with the threadpool

PR #75, opened 2020-02-06
MIGENG-417 replacing One-To-One with a Many-to-One

jonathanvila:
https://issues.redhat.com/browse/MIGENG-417

No real impact on functionality


src/main/java/org/jboss/xavier/analytics/pojo/output/workload/inventory/WorkloadInventoryReportModel.java
```diff
@@ -65,7 +65,7 @@
     )
     private Long id;
 
-    @OneToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.LAZY)
```
carlosthe19916, line 68: @jonathanvila Since you are enhancing this relation I suggest we can add `optional = false` so we are 100% sure that there won't exist any `WorkloadInventoryReportModel` without an `AnalysisModel`. this change should add `not null` constraint to the column `analysis_id` into the `workload_inventory_report_model` table.

```suggestion
    @ManyToOne(fetch = FetchType.LAZY, optional = false)
```

jonathanvila: Well, it won't make any harm, but it's totally impossible to have a null there as it has a foreign key against analysis_id where id is the primary therefore not null
    "fkt966e5qfj0hwmb0oecel9qbyw" FOREIGN KEY (analysis_id) REFERENCES analysis_model(id)

carlosthe19916: @jonathanvila Let me clarify my previous comment. Right now, we have a foreign key in `workload_inventory_report_model` which points to `analysis_model`; however, having a foreign key in `workload_inventory_report_model` does not force the table to contain not null values into the column `analysis_id`. For example you can execute these commands:

```
insert into workload_inventory_report_model(id) values (-1);
select * from workload_inventory_report_model where id=-1;
```
Screenshot from 2020-02-06 17-12-50

You will see that the previous command can be executed; it means: "I've created a workload_inventory_report_model without an analysis_model"

This is the current definition of the table, and as you can see the column `analysis_id` accept null values.

Screenshot from 2020-02-06 17-13-45

I think we missed this kind of small detail in many of our current Models; however, we can improve it by the time so we have concise model definitions.

jonathanvila: you are right.
We should revisit our whole model design .... obviously that foreign key should not be nullable
This concerns me more then, that we can find other issues in the model :(

So I would suggest not doing this small change you suggests in this PR, but open another issue of revisiting the whole model adding this and all other changes needed

carlosthe19916: Ok, let's create a ticket for analyzing all our models.

PR #76, opened 2020-02-06
MIGENG-416 setting LAZY to the relationship from AnalysisModel 

jonathanvila:
https://issues.redhat.com/browse/MIGENG-416
Changed to lazy the relationship between AnalysisModel and the list of VMWorkloadInventoryReportModel in order to reduce resources consumption

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/AnalysisModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/AnalysisModel.java
@@ -32,7 +32,7 @@ public class AnalysisModel
     )
     private Long id;
 
-    @OneToOne(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+    @OneToOne(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
     @JsonIgnore
     private InitialSavingsEstimationReportModel initialSavingsEstimationReportModel;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/output/AnalysisModel.java
```diff
@@ -32,11 +32,11 @@
     )
     private Long id;
 
-    @OneToOne(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+    @OneToOne(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
     @JsonIgnore
     private InitialSavingsEstimationReportModel initialSavingsEstimationReportModel;
 
-    @OneToMany(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.EAGER)
+    @OneToMany(mappedBy = "analysis", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
```
carlosthe19916, line 39: 1. After we change from EAGER to LAZY the deprecated method `addWorkloadInventoryReportModel` is not going to work anymore, which is expected and ok. I suggest to delete this deprecated method in this PR. https://github.com/project-xavier/xavier-integration/blob/292c2494f912d7289119b4a8af365265b117d057/src/main/java/org/jboss/xavier/integrations/jpa/service/AnalysisService.java#L69

jonathanvila: +1, if fact I dont understand why we did set it to deprecated and no other method in the code was using it.....

carlosthe19916 (approved):
This PR looks good to me.

mrizzi (commented):
I ran a test in CI comparing `master` branch performance with the one of this PR and there a nice improvement :+1: 
@mperezco was able to see his last report (`1033`) with this PR as well (previously he just received "Failed to load reports" notifications)

PR #80, opened 2020-02-14
MIGENG-429 Changed the condition to send the notification

mrizzi:
https://issues.redhat.com/browse/MIGENG-429


carlosthe19916 (approved):
Thanks @mrizzi, now we will get notifications when the daily Cron fails and also whenever a commit into the master branch fails

PR #82, opened 2020-02-14
MIGENG-429 Added specific email because ML failed

mrizzi:
https://issues.redhat.com/browse/MIGENG-429


carlosthe19916 (approved):
@mrizzi since we couldn't make it work with the mailing list, let's use custom emails until we solve the problem with our mailing list.

PR #84, opened 2020-02-21
MIGENG-358: populate new field in VMWorkloadInventoryModel

m-brophy:
https://issues.redhat.com/browse/MIGENG-358


src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
```diff
@@ -75,6 +75,15 @@ private VMWorkloadInventoryModel createVMWorkloadInventoryModel(Map vmStructMap)
             model.setHasRdmDisk(hasRdmDisk);
         }
 
+        String cpuAffinity = readValueFromExpandedEnvVarPath(CPUAFFINITYPATH, vmStructMap);
+        if (cpuAffinity != null) {
```
jonathanvila: what do you think about reducing the code to 
`model.setCpuAffinityNotNull(cpuAffinity != null)`

But, what happens if the field is set to empty "" ? does it count as != null, I mean from business pov ...

src/main/resources/application.properties
```diff
@@ -125,6 +126,7 @@ cloudforms.manifest.0.vmworkloadinventory.numCpuPath=$.ManageIQ::Providers::Vmwa
 cloudforms.manifest.0.vmworkloadinventory.numCoresPerSocketPath=$.ManageIQ::Providers::Vmware::InfraManager.hosts[*].vms[?(@.id == {id})].num_cpu
 cloudforms.manifest.0.vmworkloadinventory.hasRDMDiskPath=$.ManageIQ::Providers::Vmware::InfraManager.hosts[*].vms[?(@.id == {id})].has_rdm_disk
 cloudforms.manifest.0.vmworkloadinventory.ramSizeInBytesPath=$.ManageIQ::Providers::Vmware::InfraManager.hosts[*].vms[?(@.id == {id})].ram_size_in_bytes
+cloudforms.manifest.0.vmworkloadinventory.cpuAffinityPath=$.ManageIQ::Providers::Vmware::InfraManager.hosts[*].vms[?(@.id == {id})].cpu_affinity
```
jonathanvila: this field didnt exist on version 0

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
m-brophy: @jonathanvila , the business requirement is to raise a flag in xavier-analytics if cpuAffinity is null. So on the first point, I think that's a good idea, It's done.

m-brophy: And I've changed the code to treat "" as null

carlosthe19916 (approved):
@m-brophy Thanks for all the effort in this PR. I've tested this PR and https://github.com/project-xavier/xavier-analytics/pull/48 together in CI and everything worked fine.
- The new flag appears in the Workload Migration Summary Report (Flag table)
- The new flag is available for filtering in the Workload Inventory Report

First I'll merge https://github.com/project-xavier/xavier-analytics/pull/48 and then I'll proceed with this one.

**Note: for adding the new Flag Mappings to the DB we can use this SQL:**
`insert into flag_assessment_model(flag, os_name, flag_label, assessment) values('CPU Affinity', '', 'CPU affinity assigns specific Host CPU configuration to the VM', 'Once migrated the VM will require a CPU configuration review');`

Screenshot from 2020-03-18 12-01-35

Screenshot from 2020-03-18 12-02-13

PR #85, opened 2020-02-27
MIGENG-384: MA - Bug: VM with shared disks flag, evaluated as easy to migrate

carlosthe19916:
https://issues.redhat.com/browse/MIGENG-384

- Created new route `direct:reevaluate-workload-inventory-reports` which should be called just after the flag `Shared Disk` is added to `WorkloadInventoryReportModels`

- `direct:reevaluate-workload-inventory-reports` calls the KieServer using `KieSessionId=WorkloadInventoryComplexityKSession0` and after that update the `Complexity` field of `WorkloadInventoryReportModel`


src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_DirectCalculateFlagSharedDisksTest.java
```diff
@@ -60,18 +70,29 @@ public void mainRouteBuilder_DirectCalculate_JSONGiven_ShouldReturnExpectedCalcu
...
         String body = IOUtils.resourceToString(fileName, StandardCharsets.UTF_8, this.getClass().getClassLoader());
 
-        camelContext.createProducerTemplate().sendBodyAndHeaders("direct:flags-shared-disks", body, headers);
+        Exchange result = camelContext.createProducerTemplate().send("direct:flags-shared-disks", exchange -> {
+            exchange.getIn().setBody(body);
+            exchange.getIn().setHeaders(headers);
+        });
 
-        verify(workloadInventoryReportService).saveAll(workloadInventoryReportModels);
+        assertThat(workloadInventoryReportModels).isEqualTo(result.getIn().getBody());
```
mrizzi: Since it's `assertThat(actual).isEqualTo(expected)` in order to have the right output in case of failures, the check should be
```suggestion
        assertThat(result.getIn().getBody()).isEqualTo(workloadInventoryReportModels);
```

src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_DirectCalculateFlagSharedDisksTest.java
```diff
@@ -105,9 +126,78 @@ public void mainRouteBuilder_DirectCalculate_JSONOnVersion1_0_0Given_ShouldRetur
 
         camelContext.createProducerTemplate().sendBodyAndHeaders("direct:flags-shared-disks", body, headers);
 
+        //Then
+        Exchange result = camelContext.createProducerTemplate().send("direct:flags-shared-disks", exchange -> {
```
mrizzi, line 128: Could we invoke `send` just once and hence removing the above `sendBodyAndHeaders` method invocation?

mrizzi (commented):
@carlosthe19916 i think we talked about this change quite a lot so i've just reported some minor changes in tests

src/test/java/org/jboss/xavier/integrations/route/MainRouteBuilder_DirectCalculateFlagSharedDisksTest.java
carlosthe19916: @mrizzi Thanks for detecting this

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
```diff
@@ -72,6 +72,10 @@
         return reportRepository.findAll(specification, pageable);
     }
 
+    public WorkloadInventoryReportModel findOneById(Long id) {
```
mrizzi: I think we should try as much as we can to keep the `owner` check in place to avoid having "unsafe" method to read data.

mrizzi (commented):
Reported a concern about adding "unsafe" read method

mrizzi (approved):
Thanks a lot @carlosthe19916 for fixing this bug :+1:

PR #86, opened 2020-02-27
MIGENG-433 Fixed RecommendedTargetsIMSModel.calculateRecommendedTargetsIMS query

mrizzi:
https://issues.redhat.com/browse/MIGENG-433

- `convert2rhel` changed to `rhel` according to https://github.com/project-xavier/xavier-analytics/pull/40/files#diff-549511b11dc443fded69b06b7544a68aR60
- added full (14VMs) WIR test against updated `cfme_inventory-20190912-demolab-withssa-workload-inventory-report.json`
- updated `assertj-core` to have `usingRecursiveComparison()` method (ref https://github.com/joel-costigliola/assertj-core/commit/11cb2727fb4afc0df9fe11996dde7cecc028bd2e)

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
@@ -21,7 +21,7 @@ import javax.persistence.*;
 
 @NamedNativeQuery(
         name = "RecommendedTargetsIMSModel.calculateRecommendedTargetsIMS",
-        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='convert2rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
+        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
         resultSetMapping = "mappingRecommendedTargetsIMSModels"
 )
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
```diff
@@ -21,7 +21,7 @@
 
 @NamedNativeQuery(
         name = "RecommendedTargetsIMSModel.calculateRecommendedTargetsIMS",
-        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='convert2rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
+        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
```
m-brophy, line 24: This might sound daft, but there is also now a target of"None". Do we need to include this as well?

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
```diff
@@ -430,11 +430,18 @@ public void configure() {
             softly.assertThat(workloadInventoryReport.getBody().getContent().stream().filter(e -> e.getComplexity().contains("Unsupported")).count()).isEqualTo(1);
             softly.assertThat(workloadInventoryReport.getBody().getContent().stream().flatMap(e -> e.getRecommendedTargetsIMS().stream()).distinct().count()).isEqualTo(4);
             softly.assertThat(workloadInventoryReport.getBody().getContent().stream().filter(e -> e.getRecommendedTargetsIMS().contains("OSP")).count()).isEqualTo(11);
+            softly.assertThat(workloadInventoryReport.getBody().getContent().stream().filter(e -> e.getRecommendedTargetsIMS().contains("RHEL")).count()).isEqualTo(4);
```
m-brophy, line 433:
```suggestion
            softly.assertThat(workloadInventoryReport.getBody().getContent().stream().filter(e -> e.getRecommendedTargetsIMS().contains("RHEL")).count()).isEqualTo(4);
            softly.assertThat(workloadInventoryReport.getBody().getContent().stream().filter(e -> e.getRecommendedTargetsIMS().contains("None")).count()).isEqualTo(1);
```

We could add a check for target "None" here ( I debugged into a test run to check how many were in there).

mrizzi: The ` assertThat(workloadInventoryReport.getBody().getContent().toArray())` check below tests everything and hence also that there's just [one VM with `None` target](https://github.com/project-xavier/xavier-integration/pull/86/files#diff-e013e18be0a878695d95297e44f564f2R356).
But, since testing is never wrong and "better safe than sorry" always applies, i'll merge your check.
Thanks @m-brophy :+1:

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
mrizzi: That's a very good question @m-brophy.
I discussed with @PhilipCattanach and if the user wants to see the `None` target VMs, s/he has to use the Workload inventory Report and filter by target. 
On the Workload Summary Report we show only the data we want to show and since `None` percentage is not part of the `Target recommendation` box the `RecommendedTargetsIMSModel` bean (used to send the percentages values) should not change and hence remain consistent with the UI element it feeds with data.

m-brophy (approved):

PR #89, opened 2020-03-04
Migeng 422 used disk storage

jonathanvila:
https://issues.redhat.com/browse/MIGENG-422

If a VM.used_disk_storage field is present , it will be taken as the value for the VM, if not the previous field : VM.disks.size_on_disk will be summed into the VM

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
+++ b/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
@@ -106,13 +106,17 @@ public class VMWorkloadInventoryCalculator extends AbstractVMWorkloadInventoryCa
 
     private Long getDiskSpaceList(Map vmStructMap) {
         // If the VM.used_disk_storage is present use it, if not use VM.DISK[*].size_on_disk
-        String usedDiskStoragePath = getExpandedPath(USEDDISKSTORAGEPATH, vmStructMap);
-        Number used_disk_storage = (Number) vmStructMap.get(usedDiskStoragePath);
-        if (used_disk_storage != null) {
-            return used_disk_storage.longValue();
-        } else {
-            List<Number> hardwareDisksList = readListValuesFromExpandedEnvVarPath(HARDWAREDISKSIZEPATH, vmStructMap);
-            return hardwareDisksList.stream().filter(Objects::nonNull).mapToLong(Number::longValue).sum();
+        try {
+            String usedDiskStoragePath = getExpandedPath(USEDDISKSTORAGEPATH, vmStructMap);
+            Number used_disk_storage = (Number) vmStructMap.get(usedDiskStoragePath);
+            if (used_disk_storage != null) {
+                return used_disk_storage.longValue();
+            }
+        } catch (Exception e) {
+            // In versions previous to 1_0_0 it will fail because there is no such property
         }
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
```diff
@@ -104,4 +103,20 @@ private VMWorkloadInventoryModel createVMWorkloadInventoryModel(Map vmStructMap)
...
+    private Long getDiskSpaceList(Map vmStructMap) {
+        // If the VM.used_disk_storage is present use it, if not use VM.DISK[*].size_on_disk
+        try {
+            String usedDiskStoragePath = getExpandedPath(USEDDISKSTORAGEPATH, vmStructMap);
+            Number used_disk_storage = (Number) vmStructMap.get(usedDiskStoragePath);
+            if (used_disk_storage != null) {
+                return used_disk_storage.longValue();
+            }
+        } catch (Exception e) {
+            // In versions previous to 1_0_0 it will fail because there is no such property
```
carlosthe19916, line 116: @jonathanvila do you think it is worth adding a log here instead of not doing anything in the `catch`?

carlosthe19916 (commented):
Everything looks good. The disk size will be determined using `used_disk_storage` or `size_on_disk` according to the data available in the payload file.

I've added just a very minor question.

Screenshot from 2020-03-09 08-08-55

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/VMWorkloadInventoryCalculator.java
jonathanvila: well, we can add the log, this will only happen when we are processing a payload with a version previous to 1.0.0 version, so probably never will have that log. That's why I didnt add that log.

Do you think it's still worth it to add it ?

carlosthe19916: @jonathanvila I see... but if we never use an old version the Catch will never be executed so I think it might be worth adding a warning log to say that `xavier-integration` is using an old version

jonathanvila: +1

carlosthe19916 (approved):

carlosthe19916:
@jonathanvila We already discussed this but just to keep trace about it I'll write it here:
- We need to verify that `vm.used_disk_storage` and `vm.hardware.disks.size` have the same 
unit of measurement (KB, MB, GB) and make the conversions if needed.

carlosthe19916:
> @jonathanvila We already discussed this but just to keep trace about it I'll write it here:
> 
> * We need to verify that `vm.used_disk_storage` and `vm.hardware.disks.size` have the same
>   unit of measurement (KB, MB, GB) and make the conversions if needed.

We received confirmation that both fields have the same unit measurement. For more information see
- https://issues.redhat.com/browse/MIGENG-422?focusedCommentId=13998148&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13998148
- https://issues.redhat.com/browse/MIGENG-422?focusedCommentId=13998171&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13998171

So this PR is working fine. Thanks @jonathanvila for fixing this bug

PR #93, opened 2020-03-13
MIGENG-324: MA - Prototype RBAC - Remove IS_ORG_ADMIN header and checks

carlosthe19916:
https://issues.redhat.com/browse/MIGENG-324

- Remove `is_org_admin` header and checks from the RBAC components. Every single user should be treated like any common user.


mrizzi (approved):
Thanks @carlosthe19916 for enhancing this accordingly to the recommendations we got :+1:

PR #94, opened 2020-03-16
MIGENG-437: CNV Target - WSR - changes to Queries, Model and DB

carlosthe19916:
https://issues.redhat.com/browse/MIGENG-437

- Added CNV to `RecommendedTargetsIMSModel`
- The end2end need to be enhanced to cover the change


src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
```diff
@@ -14,14 +14,22 @@
...
+                        @ColumnResult(name = "rhel", type = Integer.class),
+                        @ColumnResult(name = "cnv", type = Integer.class)
                 }
         )
 )
 
 @NamedNativeQuery(
         name = "RecommendedTargetsIMSModel.calculateRecommendedTargetsIMS",
-        query = "select count(distinct wi.id) as total, coalesce(sum(case when lower(rt.recommended_targetsims)='rhv' then 1 else 0 end), 0) as rhv, coalesce(sum(case when lower(rt.recommended_targetsims)='osp' then 1 else 0 end), 0) as osp, coalesce(sum(case when lower(rt.recommended_targetsims)='rhel' then 1 else 0 end), 0) as rhel from workload_inventory_report_model_recommended_targetsims rt right join workload_inventory_report_model wi on rt.workload_inventory_report_model_id=wi.id where wi.analysis_id = :analysisId",
+        query = "select count(distinct wi.id) as total, \n" +
```
jonathanvila: I understand the query is divided in lines for clarity, but why including the \n in the query sent to the database ?

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
```diff
@@ -115,6 +133,7 @@ public String toString() {
                 ", rhv='" + rhv +
                 ", rhel=" + rhel +
                 ", osp=" + osp +
+                ", cnv=" + cnv +
```
jonathanvila: here we can expect that if cnv is null , this will print "null" , is that what we want ?

src/main/java/org/jboss/xavier/analytics/pojo/output/workload/summary/RecommendedTargetsIMSModel.java
carlosthe19916: Thanks @jonathanvila I'll remove the \n characters

PR #95, opened 2020-03-16
RBAC checks in all Rest endpoints

carlosthe19916:
- Added RBAC checks to all rest endpoints


m-brophy (approved):

mrizzi (approved):
Tested in CI and using Camel console i was able to see the fetch of permissions for every call done while navigating in the UI.

Thanks @carlosthe19916

PR #96, opened 2020-03-24
MIGENG-504: Unable to download report via api url

carlosthe19916:
https://issues.redhat.com/browse/MIGENG-504

Adding the "Accept Header" for fetching data from the RBAC Server.

- An error occurred when accessing the API thought the browser.
- The Reason of the error is that the Apache Camel Route is not specifying the "Accept" Http Header, therefore the route is using the came "Accept" Http Header that comes from the browser.

Notes:
- `Content-type` is a way to specify the media type of request being sent from the client to the server (valid on POST and PUT)
- `Accept` header is a way for a client to specify the media type of the response content it is expecting.

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
+++ b/src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
@@ -61,7 +61,7 @@ public class RBACRouteBuilder extends RouteBuilder {
                 .loopDoWhile(exchange -> exchange.getIn().getHeader(RBAC_NEXT_LINK) != null)
                     .setHeader(Exchange.HTTP_METHOD, constant(HttpMethods.GET))
                     .setHeader(Exchange.CONTENT_TYPE, constant(ContentType.APPLICATION_JSON.getMimeType())) // Content-type is a way to specify the media type of request being sent from the client to the server (valid on POST and PUT)
-                    .setHeader(HttpHeaders.ACCEPT).constant(ContentType.APPLICATION_JSON.getMimeType()) // Accept header is a way for a client to specify the media type of the response content it is expecting.
+                    .setHeader(HttpHeaders.ACCEPT, constant(ContentType.APPLICATION_JSON.getMimeType())) // Accept header is a way for a client to specify the media type of the response content it is expecting.
                     .setHeader(Exchange.HTTP_PATH, constant(rbacPath))
                     .process(exchange -> {
                         String httpQuery;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -58,7 +58,8 @@ public void configure() throws Exception {
                 .setHeader(RBAC_NEXT_LINK, constant(""))
                 .loopDoWhile(exchange -> exchange.getIn().getHeader(RBAC_NEXT_LINK) != null)
                     .setHeader(Exchange.HTTP_METHOD, constant(HttpMethods.GET))
-                    .setHeader(Exchange.CONTENT_TYPE, constant("application/json"))
+                    .setHeader(Exchange.CONTENT_TYPE, constant("application/json")) // Content-type is a way to specify the media type of request being sent from the client to the server (valid on POST and PUT)
```
mrizzi: I know this code is not part of this change but i figured out we're already using `ContentType` class in [MainRouteBuilder](https://github.com/project-xavier/xavier-integration/blob/master/src/main/java/org/jboss/xavier/integrations/route/MainRouteBuilder.java#L356).
```suggestion
                    .setHeader(Exchange.CONTENT_TYPE, constant(ContentType.APPLICATION_JSON)) // Content-type is a way to specify the media type of request being sent from the client to the server (valid on POST and PUT)
```

src/main/java/org/jboss/xavier/integrations/rbac/RBACRouteBuilder.java
```diff
@@ -58,7 +58,8 @@ public void configure() throws Exception {
...
+                    .setHeader("Accept").simple("application/json") // Accept header is a way for a client to specify the media type of the response content it is expecting.
```
mrizzi: Same as above with the addition of using `constant` since it's not changing.
```suggestion
                    .setHeader(HttpHeaders.ACCEPT, constant(ContentType.APPLICATION_JSON)) // Accept header is a way for a client to specify the media type of the response content it is expecting.
```

mrizzi (commented):
Added some comments on the code.
The suggested changes need also imports so please fix them as well.

Then is it possible to have a test for this change please?

carlosthe19916:
> Added some comments on the code.
> The suggested changes need also imports so please fix them as well.
> 
> Then is it possible to have a test for this change please?

Thanks for the review @mrizzi. I've solved your comments.

mrizzi (approved):
Thanks for fixing this @carlosthe19916 :+1:

PR #106, opened 2020-05-19
MIGENG-606 added dependencies on the upgrade to Fuse 7.6

jonathanvila:
https://issues.redhat.com/browse/MIGENG-606

Request is to have OpenApi v3 documentation .
 In order to achieve that we need to use `camel-rest-openapi`, and for that we need to upgrade to Fuse 7.6 that includes some backports from Camel 3 : https://access.redhat.com/documentation/en-us/red_hat_fuse/7.6/html/release_notes_for_red_hat_fuse_7.6/resolved-issues#resolved-enhancements
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,10 +14,11 @@
         <run.profiles>dev</run.profiles>
 
         <!-- configure the Fuse version you want to use here -->
-        <fuse.version>7.3.0.fuse-730058-redhat-00001</fuse.version>
+        <fuse.version>7.6.0.fuse-760027-redhat-00001</fuse.version>
 
         <drools.version>7.18.0.Final-redhat-00002</drools.version>
         <activemq.artemis.version>2.7.0</activemq.artemis.version>
+        <spring.version>1.5.22.RELEASE</spring.version>
 
         <!-- maven plugin versions -->
         <maven-compiler-plugin.version>3.6.0</maven-compiler-plugin.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -37,6 +38,14 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <!-- Import dependency management from Spring Boot -->
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-dependencies</artifactId>
+                <version>${spring.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
```
carlosthe19916: Is there a special reason for adding this block? I've verified that even without this block of code the default Spring version is `1.5.22.RELEASE` anyway. Is it a good idea to use the default version of Spring? wdyt?

carlosthe19916 (commented):
@jonathanvila I could verify the changes introduced in the PR and to me, it looks good. I added just one minor question and once answered I'll approve and merge this PR.

BEFORE:
Screenshot from 2020-05-21 11-32-01

AFTER:
Screenshot from 2020-05-21 11-32-30

pom.xml
jonathanvila: Yep, it was there because in the first changes I used org.apache.camel:camel-bom:2.21.0.fuse-760027-redhat-00001 as suggested by Fuse team, but later I changed again to fuse dependency. So you are right, that dependency on spring-boot can be removed.

carlosthe19916 (approved):

PR #109, opened 2020-05-26
Migeng 623 api req cost savings report

jonathanvila:
https://issues.redhat.com/browse/MIGENG-623


src/main/resources/spring/camel-context.xml
```diff
@@ -201,14 +216,20 @@
...
                     <!-- RBAC check start-->
                     <setHeader headerName="${type:org.jboss.xavier.integrations.rbac.RBACRouteBuilder.RBAC_ENDPOINT_RESOURCE_NAME}"><constant>analysis.initialSavingEstimation</constant></setHeader>
                     <setHeader headerName="${type:org.jboss.xavier.integrations.rbac.RBACRouteBuilder.RBAC_ENDPOINT_RESOURCE_PERMISSION}"><constant>read</constant></setHeader>
                     <to uri="direct:check-rbac-permissions" />
                     <!-- RBAC check end-->
 
                     <bean ref="initialSavingsEstimationReportService" method="findByAnalysisOwnerAndAnalysisId(${header.${type:org.jboss.xavier.integrations.route.RouteBuilderExceptionHandler.USERNAME}}, ${header.id})" />
+                    <choice>
+                        <when>
+                            <simple>${body} == null</simple>
```
carlosthe19916, line 229: @jonathanvila When we call `/report/{id}/initial-saving-estimation` there are 2 possible scenarios when `body == null` will be true:
1. When the `{id}` does not exist in the Database
2. When the `{id}` does not belong to the user (the report was created by another user).

My first thought is that we should return 2 different errors in each case; in the first case a `404` error and in the second case a `403 (forbidden)` error. However, this approach will require searching for reports regardless of who is the owner of it. What do you think? I'm ok returning `404` in the 2 scenarios I've described but I want to know your opinion.

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
```diff
@@ -70,6 +70,8 @@
 
 import javax.inject.Inject;
 import javax.inject.Named;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.NotFoundException;
```
carlosthe19916: These imports are not needed

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
```diff
@@ -182,6 +185,18 @@
     @Value("${test.bigfile.vms_expected:5254}")
     private int numberVMsExpected_InBigFile;
 
+    @Inject
+    CamelContext camelContext;
+
+    @Inject
+    AmazonS3 amazonS3;
```
carlosthe19916: I think this import is not used anywhere

carlosthe19916 (commented):
@jonathanvila I've tested this PR and it looks good to me. I've added just minor changes. Could you also please take a look to the question I've asked regarding `404` and `403` errors?

src/main/resources/spring/camel-context.xml
jonathanvila: I think that we dont have to expose our internals..... if you are checking an Id you dont have access the return of the query will be null, therefore a 404 is what I think we have to return.

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
jonathanvila: +1

src/test/java/org/jboss/xavier/integrations/EndToEndTest.java
jonathanvila: +1

carlosthe19916 (approved):

PR #110, opened 2020-05-27
MIGENG-624: OpenAPI - Workload Inventory Report

carlosthe19916:
https://issues.redhat.com/browse/MIGENG-624

changes:
- [x] Return correct Error Codes
- [x] Pagination parameters : limit,offset
- [x] Pagination maximum value for limit
- [x] Pagination response object : meta,links,data
- [x] Links object containing First and Last
- [x] Filtering results returned in the data element
- [x] Sorting response inside data element
- [ ] Date data type folling ISO-8601 format
- [ ] Objects canonical ID
- [ ] Operations MUST have an operationId defined

This PR should be deployed together with this UI PR: https://github.com/project-xavier/xavier-ui/pull/100


src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
```diff
@@ -24,6 +24,10 @@
     WorkloadInventoryReportRepository reportRepository;
 
     public static Sort getWorkloadInventoryReportModelSort(SortBean sortBean) {
```
jonathanvila: I'm concerned about static methods in a multithread system, and moreover using a Singleton (the default case with the CDI).
wdyt ?

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
```diff
@@ -68,6 +72,7 @@ public static Sort getWorkloadInventoryReportModelSort(SortBean sortBean) {
         return reportRepository.findAll(specification, sort);
     }
 
+    @Deprecated
```
jonathanvila: What is the purpose of deprecating if we are the only clients of this method ?
I would suggest to simply delete the method

src/main/java/org/jboss/xavier/integrations/route/model/PageResponse.java
```diff
@@ -0,0 +1,102 @@
+package org.jboss.xavier.integrations.route.model;
+
+import java.util.List;
+
+public class PageResponse<T> {
```
jonathanvila, line 5: Using Generics here but PageResponse&lt;Object&gt; in the ApiService..... well, do you think it would be worth to find a unified use ?

src/main/resources/spring/camel-context.xml
```diff
@@ -233,10 +232,14 @@
                     <to uri="direct:check-rbac-permissions" />
                     <!-- RBAC check end-->
 
-                    <to uri="direct:to-paginationBean"/>
-                    <to uri="direct:to-sortBean"/>
+                    <setHeader headerName="sort_by_defaults">
```
jonathanvila: I see a redundant header , as I think we could use something similar : 
<param name="sort_by" type="query" defaultValue="provider,datacenter,cluster,vmName" required="false" description="Define sort field which will be use" />
As the code is already doing a **sort_by_defaults.split**

src/main/resources/spring/camel-context.xml
```diff
@@ -233,10 +232,14 @@
...
+                        <constant>provider,datacenter,cluster,vmName</constant>
+                    </setHeader>
+                    <to uri="direct:add-pageable-header"/>
```
jonathanvila: I would see more encapsulated and responsabilities better splited if it's the Service who is responsible of creating the Pageable object, as it was before.
wdyt ?

src/main/resources/spring/camel-context.xml
carlosthe19916: this was replaced by `<param name="sort_by" type="query" defaultValue="provider,datacenter,cluster,vmName"`

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
carlosthe19916: @jonathanvila this specific method is also in the current master branch and it is not part of this PR. However, I think I can remove this method together with the deprecated one you mentioned in your other comment.

carlosthe19916: this static method was removed

src/main/java/org/jboss/xavier/integrations/route/model/PageResponse.java
carlosthe19916: @jonathanvila I don't think there is something we can do since ApiService does not have a way to know the Class the Page is using. Is there anything special you have in mind?

src/main/resources/spring/camel-context.xml
carlosthe19916: @jonathanvila I've removed `<to uri="direct:add-pageable-header"/>` so the camel routes remain the same as it is right now in the master branch. The definition and creation of the object `Pageable` is made inside the services. What changed is just the name of the fields of `PageBean`; in the past, it used `page, size` and now it uses `offset, limit`

src/main/java/org/jboss/xavier/integrations/jpa/service/WorkloadInventoryReportService.java
carlosthe19916: this method is not deprecated anymore

src/main/resources/spring/camel-context.xml
jonathanvila: +1

src/main/java/org/jboss/xavier/integrations/route/model/PageResponse.java
jonathanvila: Well, I see it's a good idea to use Generics.
So, ApiService could have a method like this 
```
     public <T> PageResponse<T> mapToCustomPage(Exchange exchange, Class<T> expectedClass) {
``` and then we would also have the real object, instead of Object.

Comment (carlosthe19916) [reply src/main/java/org/jboss/xavier/integrations/route/model/PageResponse.java:5]: @jonathanvila this would imply to change the way we call `mapToCustomPage`. For instance we should need to add an extra parameter in camel-context.xml 

```
<bean ref="apiService" method="mapToCustomPage(${exchange}, CLASS_TYPE)"/>
```

I suggest to keep the current version of ApiService so we continue calling the service using:
```
<bean ref="apiService" method="mapToCustomPage(${exchange})"/>
```

jonathanvila: why ? which is the benefit or drawback you see here ?
I would say that adding or not adding an extra param doesnt sound like a hard reason....

jonathanvila:
Test 1 :
* return proper error codes 
:heavy_check_mark: if we ask for an incorrect numeric report gives successfully a 404
:x: if we ask for an incorrect alfa report ( "ff9900" ) gives wrongly a 200 ""

* pagination parameters : limit, offset
:heavy_check_mark: OK

* pagination maximum value
:x: returns a number with a coma "1,000"

* pagination response object
:x: param "previous" should be "prev"

* links objects containing first, last, and may contain next, prev
:heavy_check_mark: OK

* filtering results in data element
:no_entry:  not applicable

* sorting results in data element
:heavy_check_mark: OK

* date type ISO
:heavy_check_mark: OK

* objects canonical id
ON HOLD

* operations operationID
:heavy_check_mark: OK

jonathanvila:
Test 2 :

-  return proper error codes
    :heavy_check_mark: if we ask for an incorrect numeric report gives successfully a 404
    :heavy_check_mark: if we ask for an incorrect alfa report ( "ff9900" ) gives a 500
-  pagination parameters : limit, offset
     :heavy_check_mark: OK
-  pagination maximum value
     :heavy_check_mark: OK
-  pagination response object
     :heavy_check_mark: OK
-  links objects containing first, last, and may contain next, prev
     :heavy_check_mark: OK
-  filtering results in data element
    :no_entry: not applicable
-   sorting results in data element
     :heavy_check_mark: OK

    date type ISO
     :heavy_check_mark: OK

    objects canonical id
    ON HOLD

    operations operationID
     :heavy_check_mark: OK

LGTM :+1:

jonathanvila (approved):

PR #112, opened 2020-06-04
Refactor end2end

jonathanvila:
Refactoring EndToEndTest in order to be able to have individual tests and a unique point of infra creation with TestContainers

Things to consider :
* tests are not order guaranteed so we need to think on the tests as in isolation
* every test needs to check the condition without expecting other tests have populated info


jonathanvila:
retest

PR #113, opened 2020-06-04
MIGENG-636 reduced time on Stress

jonathanvila:


carlosthe19916 (approved):

carlosthe19916:
Thanks @jonathanvila for the enhancement reducing the time of execution of our end2end test and at the same time keep the purpose and logic of the tests

PR #116, opened 2020-06-19
[RFR] Fix iqe plugin name

ganeshhubale:
Signed-off-by: Ganesh Hubale <ghubale@redhat.com>

- Plugin name should match package name - https://gitlab.cee.redhat.com/insights-qe/iqe-migration-analytics-plugin/-/blob/master/iqe_migration_analytics/__init__.py#L28


mrizzi (approved):
thanks @ganeshhubale

PR #117, opened 2020-06-24
(stable) MIGENG-449 added the delete action of the S3 file and the test in End…

mrizzi:
…… (#103)

* MIGENG-449 added the delete action of the S3 file and the test in End2EndTest

* MIGENG-449 fixed URL for the delete test

* MIGENG-449 changed the way to pass the File KEY to S3
change expected status code on the test

* MIGENG-449 added arg to travis pipeline to reduce logging
removed the replacement for the S3 object key

* MIGENG-449 removed s3 component on the Delete unit test to not modify the http response code

* MIGENG-449 Added service to retrieve objects in S3
added call to get number of objects in S3 after upload and after delete
refactored hard coded constants to use proper constants

* MIGENG-449 rollback changes to retrieve S3 number of objects as AWS only gives up to 1000, will do with command line to check

* MIGENG-449 deleted unused class StorageService

* MIGENG-449 removed old code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ jobs:
         - LOGGING_LEVEL_ROOT=info
         - ANALYSIS_DATAINTEGRITY_LOG=false
       script:
-        - mvn test -Pcoverage
+        - mvn test -Pcoverage -Danalysis.dataintegrity.log=false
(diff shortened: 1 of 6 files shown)


mrizzi:
Test PR for the e2e-deploy smoke test build

PR #119, opened 2020-06-24
MIGENG-599: New field in admin export csv

m-brophy:
https://issues.redhat.com/browse/MIGENG-599

### DIFF
--- a/src/main/java/org/jboss/xavier/analytics/pojo/AdministrationMetricsModel.java
+++ b/src/main/java/org/jboss/xavier/analytics/pojo/AdministrationMetricsModel.java
@@ -14,7 +14,7 @@ public class AdministrationMetricsModel {
     @DataField(pos = 2, columnName = "Domain")
     private String ownerDomain;
 
-    @DataField(pos = 3, columnName = "Owner Account Number")
+    @DataField(pos = 3, columnName = "Owner account number")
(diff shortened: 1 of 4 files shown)


src/main/java/org/jboss/xavier/analytics/pojo/AdministrationMetricsModel.java
```diff
@@ -14,16 +14,19 @@
     @DataField(pos = 2, columnName = "Domain")
     private String ownerDomain;
 
-    @DataField(pos = 3, columnName = "Payload name")
+    @DataField(pos = 3, columnName = "Owner Account Number")
```
carlosthe19916:
```suggestion
    @DataField(pos = 3, columnName = "Owner account number")
```
After worked on https://issues.redhat.com/browse/MIGENG-654 I've learned that we should try to follow the "Title case" pattern in every component that is shown to the user, so we should only have the first letter in upper case.

carlosthe19916 (commented):
@m-brophy This PR looks good to me, I've just added a comment suggesting to change the column name of the CSV file. I've also tested it in CI and it worked fine. Once the comment is solved and Travis finishes successfully I'll be ready to merge this PR

carlosthe19916 (approved):

PR #126, opened 2020-07-15
MIGENG-673 replaced the AMQ usage by a seda pattern

jonathanvila:
https://issues.redhat.com/browse/MIGENG-673

Replacing the usage of AMQ by a `seda` Camel pattern

### DIFF
--- a/src/main/resources/spring/camel-context.xml
+++ b/src/main/resources/spring/camel-context.xml
@@ -10,16 +10,6 @@
 
     <jpa:repositories base-package="org.jboss.xavier.integrations.jpa" />
 
-    <!-- setup ActiveMQ Artemis connection factory -->
-    <bean id="artemisConnectionFactory" class="org.apache.activemq.artemis.jms.client.ActiveMQJMSConnectionFactory">
-        <constructor-arg index="0" value="tcp://${amq.server}:${amq.port:61616}"/>
-    </bean>
-
-    <bean id="jms" class="org.apache.camel.component.jms.JmsComponent">
-        <!-- specify connection factory on JMS component -->
-        <property name="connectionFactory" ref="artemisConnectionFactory"/>
-    </bean>
-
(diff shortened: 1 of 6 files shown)


jonathanvila:
retest

src/main/resources/spring/camel-context.xml
```diff
@@ -542,19 +532,8 @@
...
-            <bean method="createMigrationAnalyticsCommand(${body})" ref="decisionServerHelper"/>
-            <to id="decisionserver" uri="direct:decisionserver"/>
-            <bean id="route-extract-reports" method="extractReports" ref="decisionServerHelper"/>
-            <log id="route-log-totalPrice" message="totalPrice = ${body.totalPrice}"/>
-            <to uri="jpa:org.jboss.xavier.integrations.migrationanalytics.output.ReportDataModel" />
-        </route>
-
         <route id="route-ma">
-            <from uri="jms:queue:uploadFormInputDataModel" />
+            <from uri="seda:uploadFormInputDataModel?concurrentConsumers={{thread.concurrentConsumers:5}}"/>
```
carlosthe19916, line 536: @jonathanvila I think it should be possible to use `direct` instead of `seda`, is there a special reason to use `seda` here?... My point is that we don't need to request an additional Thread to process this piece of code and we can let the original Thread to execute it without blocking and additional Thread. Maybe I'm missing something, what do you think?

jonathanvila: Here I put the Calls diagram of xavier-integration : https://drive.google.com/file/d/1jBLCldaxChdkcGgXgwtQJFFlLltyfaVb/view?usp=sharing
In my opinion having an asynchronous call to delegate the flow is always beneficial if it's easy to handle.

carlosthe19916:
I've created https://issues.redhat.com/browse/MIGENG-696 in order to report the current bug of having async calls on `direct:send-costsavings`. The async way of executing was there even with the `amq (now seda)` route so we will fix this bug in another PR.

carlosthe19916 (approved):

PR #140, opened 2020-12-02
MIGENG-687 remove reevaluate route

m-brophy:
https://issues.redhat.com/browse/MIGENG-687?focusedCommentId=15505116&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15505116

### DIFF
--- a/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/FlagSharedDisksCalculator.java
+++ b/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/FlagSharedDisksCalculator.java
@@ -51,7 +51,6 @@ public class FlagSharedDisksCalculator extends AbstractVMWorkloadInventoryCalcul
             });
         final Set<String> vmNamesWithSharedDisk = new HashSet();
         fileNamesInVms.values().stream().filter(set -> set.size() > 1).forEach(set -> vmNamesWithSharedDisk.addAll(set));
-        vmNamesWithSharedDisk.forEach(vmname -> System.out.println("FlagSharedDisksCalculator - Shared Disk VM Name: " + vmname));
         return vmNamesWithSharedDisk;
     }
 
--- a/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/FlagSharedDisksCalculator.java
+++ b/src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/FlagSharedDisksCalculator.java
@@ -51,6 +51,7 @@ public class FlagSharedDisksCalculator extends AbstractVMWorkloadInventoryCalcul
             });
         final Set<String> vmNamesWithSharedDisk = new HashSet();
         fileNamesInVms.values().stream().filter(set -> set.size() > 1).forEach(set -> vmNamesWithSharedDisk.addAll(set));
+        vmNamesWithSharedDisk.forEach(vmname -> System.out.println("FlagSharedDisksCalculator - Shared Disk VM Name: " + vmname));
         return vmNamesWithSharedDisk;
     }
 
--- a/src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
+++ b/src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
@@ -61,7 +61,7 @@ public class VMWorkloadInventoryRoutes extends RouteBuilderExceptionHandler {
                 Set<String> vmNamesWithSharedDisk = exchange.getIn().getBody(Set.class);
                 vmNamesWithSharedDisk.forEach(vmname -> System.out.println("flags-shared-disks.settingHeader - Shared Disk VM Name: " + vmname));
                 exchange.getIn().setHeader("vmNamesWithSharedDisk", vmNamesWithSharedDisk);
-            });
+            }).to("direct:calculate-vmworkloadinventory");
 
 
     }
--- a/src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
+++ b/src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
@@ -42,7 +42,7 @@ public class VMWorkloadInventoryRoutes extends RouteBuilderExceptionHandler {
                         return vmWorkloadInventoryModel;
                     }).collect(Collectors.toList());
                     exchange.getIn().setBody(allVms);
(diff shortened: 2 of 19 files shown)


src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/FlagSharedDisksCalculator.java
```diff
@@ -51,6 +51,7 @@
             });
         final Set<String> vmNamesWithSharedDisk = new HashSet();
         fileNamesInVms.values().stream().filter(set -> set.size() > 1).forEach(set -> vmNamesWithSharedDisk.addAll(set));
+        vmNamesWithSharedDisk.forEach(vmname -> System.out.println("FlagSharedDisksCalculator - Shared Disk VM Name: " + vmname));
```
carlosthe19916: Could we remove this line? it is just printing and not doing some logic.

src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
```diff
@@ -41,42 +58,18 @@ public void configure() throws Exception {
...
-                        Long.parseLong(exchange.getIn().getHeader(MA_METADATA, Map.class).get(ANALYSIS_ID).toString()));
-                List<WorkloadInventoryReportModel> workloadInventoryReportModelsToUpdate = workloadInventoryReportModels.stream()
-                    .filter(workloadInventoryReportModel -> vmNamesWithSharedDisk.contains(workloadInventoryReportModel.getVmName()))
-                    .peek(workloadInventoryReportModel -> workloadInventoryReportModel.addFlagIMS("Shared Disk")).collect(Collectors.toList());
-                exchange.getIn().setBody(workloadInventoryReportModelsToUpdate);
-            })
-            .to("direct:reevaluate-workload-inventory-reports");
+                exchange.getIn().setHeader("vmNamesWithSharedDisk", vmNamesWithSharedDisk);
+                exchange.getIn().setBody(originalBody);
+            }).to("direct:calculate-vmworkloadinventory");
```
carlosthe19916, line 70: Maybe I'm missing something but the code inside `direct:flags-shared-disks` is not persisted or used anywhere, not even in the next route `direct:calculate-vmworkloadinventory`. I've just created a PR to your PR https://github.com/m-brophy/xavier-integration/pull/2 to be more clear what I mean. I've tested my PR in CI and I have the same results as with your PR. Could you please take a look at my PR and verify if removing this block of code does not affect the logic?

src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
```diff
@@ -28,8 +30,23 @@ public void configure() throws Exception {
...
+                    Collection<VMWorkloadInventoryModel> originalAndAmendedVms = allVms.stream().map(vmWorkloadInventoryModel ->
+                    {
+                        if (vmNamesWithSharedDisk != null && vmNamesWithSharedDisk.contains(vmWorkloadInventoryModel.getVmName())) {
+                            vmWorkloadInventoryModel.setHasSharedVmdk(true);
+                        }
+                        return vmWorkloadInventoryModel;
+                    }).collect(Collectors.toList());
+                    exchange.getIn().setBody(allVms);
+                    exchange.getIn().removeHeader("vmNamesWithSharedDisk");
+                    exchange.getIn().removeHeader("originalBody");
```
carlosthe19916: I don't see this block of code is adding value while analyzing it against xavier-analytics. Please see my comment above for more info.

carlosthe19916 (changes requested):
@m-brophy I've tested this PR in CI and it works fine; however, I still think some pieces of code are not needed anymore, could you please take a look at my comments below? maybe I'm missing something.

src/main/java/org/jboss/xavier/integrations/migrationanalytics/business/FlagSharedDisksCalculator.java
m-brophy: agreed - done

src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
m-brophy: The names of vms which use a shared disk are stored in the header which is then retrieved in the next route and used to determine which vms have .setHasSharedVmdk(true) applied to them. 

To test this block of code you need to use a payload which would have triggered the old SHARED_DISK flag via the removed reevaulate route and ensure you still get shared_vmdk flags for the appropriate vms even though there is no use_shared_vmdk: true element in the payload

src/main/java/org/jboss/xavier/integrations/route/VMWorkloadInventoryRoutes.java
```diff
@@ -28,8 +30,23 @@ public void configure() throws Exception {
...
+                .process(exchange -> {
+                    Set<String> vmNamesWithSharedDisk = exchange.getIn().getHeader("vmNamesWithSharedDisk", Set.class);
+                    Collection<VMWorkloadInventoryModel>  allVms = exchange.getIn().getBody(Collection.class);
+                    Collection<VMWorkloadInventoryModel> originalAndAmendedVms = allVms.stream().map(vmWorkloadInventoryModel ->
+                    {
+                        if (vmNamesWithSharedDisk != null && vmNamesWithSharedDisk.contains(vmWorkloadInventoryModel.getVmName())) {
+                            vmWorkloadInventoryModel.setHasSharedVmdk(true);
+                        }
+                        return vmWorkloadInventoryModel;
+                    }).collect(Collectors.toList());
```
carlosthe19916:
```suggestion
                    allVms.forEach(vmWorkloadInventoryModel ->
                    {
                        if (vmNamesWithSharedDisk != null && vmNamesWithSharedDisk.contains(vmWorkloadInventoryModel.getVmName())) {
                            vmWorkloadInventoryModel.setHasSharedVmdk(true);
                        }
                    });
```

@m-brophy this just a suggestion, it seems to me that `originalAndAmendedVms` is not used so better to use a `forEach()` instead of `map()` since we only need to iterate the collection `allVms`.

carlosthe19916 (commented):
@m-brophy I've tested this PR in CI
- Tested an old payload: It does generate the flag `shared vmdk`
- Tested a new payload: it does generate the  flag `shared vmdk`

So this PR will keep making work old payloads and also new ones. I just added a very small suggestion in the code, could you please take a look?

